PR #1, opened 2019-08-14
multiple project release support

evgeni:


settings
```diff
@@ -1,24 +1,34 @@
-VERSION="1.23"
+FOREMAN_VERSION="1.23"
```
ekohl: I've never felt very good about hardcoding this in the file. Wondering if this should also be read similar to `$PROJECT`.

settings
```diff
@@ -1,24 +1,34 @@
...
+KATELLO_VERSION="3.13"
```
ekohl: Should this be part of foreman's `SETTINGS` since they're quite tied together? Especially since we've started to share GPG keys now.

evgeni:
@ekohl have a look at my second commit :)

settings
```diff
@@ -1,32 +1,46 @@
...
+PASS_NAME_GPG="theforeman/releases/foreman/$FOREMAN_VERSION-gpg"
+PASS_NAME_KEY="theforeman/releases/foreman/$FOREMAN_VERSION-key"
 XARGS_JOBS="-n 20 -P 4"
 GIT_DIR="$HOME/dev"
 PACKAGING_DIR="$GIT_DIR/foreman-packaging"
 PACKAGING_GIT_REMOTE=origin
 TAR_PROJECTS="foreman foreman-proxy foreman-installer foreman-selinux"
 RPM_PACKAGES="foreman foreman-installer foreman-proxy foreman-release foreman-selinux"
-ROR_BASE_TAG="foreman-rails-$VERSION"
+ROR_BASE_TAG="foreman-rails-$FOREMAN_VERSION"
```
ekohl: This and RPM_PACKAGES are a bit of a foreman specific thing. Should we also have a `releases/$PROJECT/settings` file to hold these as a layered default? Then we can set it to empty in Katello.

evgeni: done

settings
evgeni: not sure what you mean here

ekohl: We currently have `releases/foreman/x.y/settings` and `releases/katello/a.b/settings`. However, could we have a `KATELLO_VERSION` in `releases/x.y/settings` and account for this in the scripts? Longer term I'd prefer for katello tags to move to `foreman-katello-x.y-$os` rather than `katello-a.b-$os` which would remove a lot of the specifics again.

This doesn't have to be done in this PR, but I do want to get your thoughts on the direction.

download_rpms
```diff
@@ -8,14 +8,20 @@ cd "$RPMDIR"
...
 for os in $OSES; do
-  tags="foreman-$VERSION-$os"
-  [[ -n $ROR_BASE_TAG ]] && tags="$tags $ROR_BASE_TAG-$os"
-  [[ $os =~ rhel ]] && tags="$tags foreman-$VERSION-nonscl-$os"
+  tags="$PROJECT-$VERSION-$os"
+  if [[ $PROJECT = foreman ]]; then
+    [[ -n $ROR_BASE_TAG ]] && tags="$tags $ROR_BASE_TAG-$os"
+    [[ $os =~ rhel ]] && tags="$tags foreman-$VERSION-nonscl-$os"
+  elif [[ $PROJECT = katello ]]; then
+    [[ $os =~ rhel ]] && tags="$tags katello-$VERSION-thirdparty-candlepin-$os katello-$VERSION-thirdparty-candlepin-$os"
```
ekohl: You're repeating the tag here
```suggestion
    [[ $os =~ rhel ]] && tags="$tags katello-$VERSION-thirdparty-candlepin-$os"
```

download_rpms
```diff
@@ -8,14 +8,20 @@ cd "$RPMDIR"
 tags=""
 
 for os in $OSES; do
-  tags="foreman-$VERSION-$os"
-  [[ -n $ROR_BASE_TAG ]] && tags="$tags $ROR_BASE_TAG-$os"
-  [[ $os =~ rhel ]] && tags="$tags foreman-$VERSION-nonscl-$os"
+  tags="$PROJECT-$VERSION-$os"
+  if [[ $PROJECT = foreman ]]; then
```
ekohl: Bash comparisons can use `==`
```suggestion
  if [[ $PROJECT == foreman ]]; then
```

download_rpms
```diff
@@ -8,14 +8,20 @@ cd "$RPMDIR"
...
+    [[ -n $ROR_BASE_TAG ]] && tags="$tags $ROR_BASE_TAG-$os"
+    [[ $os =~ rhel ]] && tags="$tags foreman-$VERSION-nonscl-$os"
```
ekohl: I'm wondering if we can have a `HAS_NONSCL=1` variable:
```suggestion
    [[ $HAS_NONSCL == 1 ]] && [[ $os =~ rhel ]] && tags="$tags $PROJECT-$VERSION-nonscl-$os"
```

Alternative implementation:

```bash
# Foreman
ROR_BASE_TAG=...
ADDITIONAL_RHEL_TAG=nonscl
# Katello
ROR_BASE_TAG=""
ADDITIONAL_RHEL_TAG=thirdparty-candlepin
```

Then
```bash
for os in $OSES; do
   tags="$PROJECT-$VERSION-$os" 
   [[ -n $ROR_BASE_TAG ]] && tags="$tags $ROR_BASE_TAG-$os" 
   [[ $os =~ rhel ]] && tags="$tags $PROJECT-$VERSION-$ADDITIONAL_RHEL_TAG-$os" 
done
```

evgeni: Why do we need this complexity anyways? Why not having a `ADDITIONAL_TAGS` (in the version specific settings) and iterate over them?

```bash
# Foreman
ADDITIONAL_TAGS="foreman-rails-$VERSION foreman-$VERSION-nonscl"

# Katello
ADDITIONAL_TAGS="katello-$VERSION-thirdparty-candlepin"

for os in $OSES; do
  tags="$PROJECT-$VERSION-$os"
  for at in ${ADDITIONAL_TAGS}; do
    tags="$tags $at-$os"
  done
done
```

Or even better, non-version-specific:

```bash
# Foreman
ADDITIONAL_TAGS="foreman-rails-@VERSION@ foreman-@VERSION@-nonscl"

# Katello
ADDITIONAL_TAGS="katello-@VERSION@-thirdparty-candlepin"

for os in $OSES; do
  additional_versioned_tags=$(echo ${ADDITIONAL_TAGS} | sed "s/@VERSION@/$VERSION/g")
  tags="$PROJECT-$VERSION-$os"
  for at in ${additional_versioned_tags}; do
    tags="$tags $at-$os"
  done
done
```

ekohl: My issue was that templating part. I didn't think of pulling in sed. That does look like a much cleaner solution. If you're already doing that, wouldn't you want to go all the way and also make the os part of the template and just make it `TAGS`?

evgeni: I made server_tags and client_tags now. it generates:

```
$ ./download_rpms
foreman-1.23-rhel7 foreman-1.23-nonscl-rhel7 foreman-rails-1.23-rhel7 foreman-client-1.23-rhel5 foreman-client-1.23-rhel6 foreman-client-1.23-rhel7 foreman-client-1.23-sles11 foreman-client-1.23-sles12 foreman-client-1.23-fedora28 foreman-client-1.23-fedora29

$ PROJECT=katello VERSION=3.13 ./download_rpms                   
katello-3.13-rhel7 katello-3.13-thirdparty-candlepin-rhel7
```

so looks correct to me?

evgeni:
does that happen often? like do we need to release client w/o server?

ekohl:
Technically they're distinct release "cycles". It would also avoid the need for `CLIENT_OSES` and `CLIENT_TAGS`. There'd just be `OSES` and `TAGS`.

evgeni:
done

evgeni:
renamed the PR, as it's not really about katello anymore ;)

evgeni:
@ekohl done

settings
```diff
@@ -1,32 +1,51 @@
...
+fi
+
+if [[ $PROJECT != foreman ]]; then
+  . "$RELEASEDIR/settings"
+  if [[ -z "$FOREMAN_VERSION" ]]; then
+    echo "please set FOREMAN_VERSION in $RELEASEDIR/settings"
+    exit 1
+  fi
+else
+  FOREMAN_VERSION="$VERSION"
```
evgeni: given we share data with foreman, I wonder if we should load the foreman settings here? esp the key definitions are duplicated now

ekohl: That was exactly what I was thinking.

releases/client/settings
```diff
@@ -0,0 +1,2 @@
+OSES="rhel5 rhel6 rhel7 sles11 sles12 fedora28 fedora29"
```
ekohl: Should we explicitly copy this so we don't have to when we change 1.24 (or later) to drop F28?

evgeni: you mean to `releases/client/1.23/settings`? probably yes.

ekohl: Exactly.

evgeni:
@ekohl ready for another round, now even with docs :D

PR #2, opened 2019-08-14
don't use koji from git for signing

evgeni:


ekohl:
I needed this in the past because koji in Fedora was broken. Perhaps it works again since Fedora 30? If you successfully used this then I'm fine with merging.

evgeni:
My laptop still runs F29 (lazy me) and it works just fine here.

PR #3, opened 2019-08-14
use find/basename to find RPMs to upload

evgeni:


ekohl:
Looking a bit closer: have you tested this? Looking at this I wonder if the relative path works without the cd. basename strips away the directory not just the extension.

evgeni:
yes. the RPMs are not *really* uploaded, they are just *written* on koji.

ekohl:
Right, so we should rename the script to write_signed_rpms to be clearer.

evgeni:
`<ewoud> naming is hard`

evgeni:
We could also pass `--write` to `import-sigs`, then this step wouldn't be needed at all anymore?

ekohl:
Honestly, for this part I mostly used what was handed over to me.

image

PR #4, opened 2019-10-21
Generalize the RPM bumping code

ekohl:
Note this is untested. It's taken from https://github.com/theforeman/foreman-packaging/blob/189fe820b1434174349d6bd6a6febecd541ab0b8/bump_rpm.sh#L53

### DIFF
--- a/bump_rpm_packaging
+++ b/bump_rpm_packaging
@@ -12,7 +12,10 @@ git checkout -b rpm/release-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
 for package in $RPM_PACKAGES ; do
 	(
 		cd packages/foreman/$package
+
 		spec_file="${package}.spec"
+		spectool --list-files $spec_file | cut -d' ' -f2 | grep http | xargs --no-run-if-empty -n 1 basename | xargs --no-run-if-empty git rm
+
 		if [[ $FULLVERSION == *-RC* ]] ; then
 			if [[ FULLVERSION == *-RC1 ]] ; then
 				echo "TODO: bump release"
@@ -21,15 +24,10 @@ for package in $RPM_PACKAGES ; do
 		else
 			sed -i 's/^\(Version:\s\+\).\+$/\1'$FULLVERSION'/ ; s/^\(%global release\s\+\).\+$/\11/ ; /^\(%global prerelease\s\)/d' $spec_file
 		fi
-		git add $spec_file
-		if [[ $package != foreman-release ]] ; then
-			git rm --ignore-unmatch --force *.tar.bz2
-			spectool -g $spec_file
-			git annex add *.tar.bz2
-		fi
 	)
 done
 
+obal source $RPM_PACKAGES
 obal changelog --message "- Release $FULLVERSION" $RPM_PACKAGES
 for package in $RPM_PACKAGES ; do
 	git add packages/foreman/$package/$package.spec
--- a/bump_rpm_packaging
+++ b/bump_rpm_packaging
@@ -12,18 +12,19 @@ git checkout -b rpm/release-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
 for package in $RPM_PACKAGES ; do
 	(
 		cd packages/foreman/$package
+		spec_file="${package}.spec"
 		if [[ $FULLVERSION == *-RC* ]] ; then
 			if [[ FULLVERSION == *-RC1 ]] ; then
 				echo "TODO: bump release"
 			fi
-			sed -i 's/^\(Version:\s\+\).\+$/\1'${FULLVERSION%%-RC*}'/ ; s/^\(%global prerelease\s\+\).\+$/\1'${FULLVERSION##*-}'/' $package.spec
+			sed -i 's/^\(Version:\s\+\).\+$/\1'${FULLVERSION%%-RC*}'/ ; s/^\(%global prerelease\s\+\).\+$/\1'${FULLVERSION##*-}'/' $spec_file
 		else
-			sed -i 's/^\(Version:\s\+\).\+$/\1'$FULLVERSION'/ ; s/^\(%global release\s\+\).\+$/\11/ ; /^\(%global prerelease\s\)/d' $package.spec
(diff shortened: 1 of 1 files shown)


bump_rpm_packaging
```diff
@@ -12,20 +12,18 @@ git checkout -b rpm/release-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
 for package in $RPM_PACKAGES ; do
 	(
 		cd packages/foreman/$package
+		spec_file=$package.spec
```
evgeni:
```suggestion
		spec_file="${package}.spec"
```

bump_rpm_packaging
```diff
@@ -12,20 +12,18 @@ git checkout -b rpm/release-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
...
-		fi
-		git add $package.spec
-		if [[ $package != foreman-release ]] ; then
-			git rm --ignore-unmatch --force *.tar.bz2
-			spectool -g $package.spec
-			git annex add *.tar.bz2
+			sed -i 's/^\(Version:\s\+\).\+$/\1'$FULLVERSION'/ ; s/^\(%global release\s\+\).\+$/\11/ ; /^\(%global prerelease\s\)/d' $spec_file
 		fi
+		git add $spec_file
+		spectool --list-files $spec_file | cut -d' ' -f2 | grep http | xargs --no-run-if-empty -n 1 basename | xargs --no-run-if-empty git rm	spectool --get-files $spec_file
```
evgeni: this line looks broken at the end

ekohl: Interesting. Didn't show on my terminal but that was just because of the width.

ekohl:
Updated

bump_rpm_packaging
```diff
@@ -12,20 +12,19 @@ git checkout -b rpm/release-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
...
-		if [[ $package != foreman-release ]] ; then
-			git rm --ignore-unmatch --force *.tar.bz2
-			spectool -g $package.spec
-			git annex add *.tar.bz2
+			sed -i 's/^\(Version:\s\+\).\+$/\1'$FULLVERSION'/ ; s/^\(%global release\s\+\).\+$/\11/ ; /^\(%global prerelease\s\)/d' $spec_file
 		fi
+		git add $spec_file
+		spectool --list-files $spec_file | cut -d' ' -f2 | grep http | xargs --no-run-if-empty -n 1 basename | xargs --no-run-if-empty git rm
+		spectool --get-files $spec_file
+		spectool --list-files $spec_file | cut -d' ' -f2 | grep http | xargs --no-run-if-empty -n 1 basename | xargs --no-run-if-empty git annex add
```
evgeni: these two lines can be replaced by `obal source ${package}`

bump_rpm_packaging
```diff
@@ -12,20 +12,19 @@ git checkout -b rpm/release-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
...
-		fi
-		git add $package.spec
-		if [[ $package != foreman-release ]] ; then
-			git rm --ignore-unmatch --force *.tar.bz2
-			spectool -g $package.spec
-			git annex add *.tar.bz2
+			sed -i 's/^\(Version:\s\+\).\+$/\1'$FULLVERSION'/ ; s/^\(%global release\s\+\).\+$/\11/ ; /^\(%global prerelease\s\)/d' $spec_file
 		fi
+		git add $spec_file
+		spectool --list-files $spec_file | cut -d' ' -f2 | grep http | xargs --no-run-if-empty -n 1 basename | xargs --no-run-if-empty git rm
```
evgeni: this needs to happen *before* you change the version in the spec file, as otherwise it will `git rm` an non-existing file ;)

bump_rpm_packaging
```diff
@@ -12,20 +12,19 @@ git checkout -b rpm/release-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
 for package in $RPM_PACKAGES ; do
 	(
 		cd packages/foreman/$package
+		spec_file="${package}.spec"
 		if [[ $FULLVERSION == *-RC* ]] ; then
 			if [[ FULLVERSION == *-RC1 ]] ; then
 				echo "TODO: bump release"
 			fi
-			sed -i 's/^\(Version:\s\+\).\+$/\1'${FULLVERSION%%-RC*}'/ ; s/^\(%global prerelease\s\+\).\+$/\1'${FULLVERSION##*-}'/' $package.spec
+			sed -i 's/^\(Version:\s\+\).\+$/\1'${FULLVERSION%%-RC*}'/ ; s/^\(%global prerelease\s\+\).\+$/\1'${FULLVERSION##*-}'/' $spec_file
```
evgeni: can we make `obal update` do this for us?

ekohl: Maybe, but let's not block this on functionality needed in obal.

bump_rpm_packaging
ekohl: Does obal source also do the git-annex? That role looks so much more complex.

evgeni: Yes it does.

bump_rpm_packaging
evgeni: Agreed.

ekohl:
Updated

bump_rpm_packaging
```diff
@@ -12,23 +12,22 @@ git checkout -b rpm/release-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
 for package in $RPM_PACKAGES ; do
 	(
 		cd packages/foreman/$package
+
+		spec_file="${package}.spec"
+		spectool --list-files $spec_file | cut -d' ' -f2 | grep http | xargs --no-run-if-empty -n 1 basename | xargs --no-run-if-empty git rm
+
 		if [[ $FULLVERSION == *-RC* ]] ; then
 			if [[ FULLVERSION == *-RC1 ]] ; then
 				echo "TODO: bump release"
```
ekohl: Side note: if we used `rc1` instead of `RC1` it would lexically sort and the release bump is not needed. However, `RC1 < nightly`.

PR #6, opened 2019-10-24
don't hardcode packages/foreman in bump_rpm

evgeni:


bump_rpm_packaging
```diff
@@ -30,7 +30,7 @@ done
 obal source $RPM_PACKAGES
 obal changelog --message "- Release $FULLVERSION" $RPM_PACKAGES
 for package in $RPM_PACKAGES ; do
-	git add packages/foreman/$package/$package.spec
+	git ${PACKAGING_SUBDIR}/$package/$package.spec
```
ekohl:
```suggestion
	git add ${PACKAGING_SUBDIR}/$package/$package.spec
```

evgeni: :see_no_evil:

PR #10, opened 2019-10-28
use obal update

evgeni:


bump_rpm_packaging
```diff
@@ -11,30 +11,18 @@ git checkout -b rpm/release-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
 
 for package in $RPM_PACKAGES ; do
```
ekohl: Given obal can handle multiple packages, should this just call `obal update` on all packages? They match in version. Also wondering if obal can increase the release on RC1.

evgeni: Probably.
And rc1 handling need to be implemented, but yes.

ekohl: Looking at the source of `fetch_sources` it always sets the release to `'1'` when it's not defined which probably breaks a lot of our workflows. When going to RC1 we must not reset.
https://github.com/theforeman/obal/blob/0d2ca6d9d001e4af14087d317cbc64d49479cb27/obal/data/roles/fetch_sources/tasks/main.yml#L77-L82

evgeni: https://github.com/theforeman/obal/pull/217

PR #13, opened 2019-10-29
Autodetect the short GPG key

ekohl:
This is only used by koji and can be automatically derived from the full key.


generate_gpg
```diff
@@ -28,5 +28,5 @@ cat > $GPG_SETTINGS <<EOF
 EOF
 gpg2 --homedir $KEYDIR --batch --gen-key $GPG_SETTINGS
 
-echo "Now set GPGKEY and FULLGPGKEY in releases/$VERSION/settings"
+echo "Now set FULLGPGKEY in releases/$VERSION/settings"
 gpg2 --homedir "$KEYDIR" --list-key "$SIGNER"
```
ekohl: Future improvement would be to automatically parse this output and append it to the settings.

PR #14, opened 2019-10-29
Add the 1.24 release

zjhuntin:


releases/foreman/1.24/settings
```diff
@@ -0,0 +1,3 @@
+FULLVERSION="1.24.0"
```
evgeni: I guess this should be
```suggestion
FULLVERSION="1.24.0-RC1"
```

releases/foreman/1.24/settings
```diff
@@ -0,0 +1,3 @@
...
+FULLGPGKEY="D885D3D9B12FA857CDECE3F1B76498E45AA9BCAD"
+GPGKEY="5aa9bcad"
```
evgeni: this one is not needed anymore after #13

zjhuntin:
@evgeni updated!

PR #16, opened 2019-11-05
katello 3.14 and katello improvements

evgeni:


bump_rpm_packaging
```diff
@@ -13,9 +13,14 @@ if [[ $FULLVERSION == *-RC* ]] ; then
 	if [[ FULLVERSION == *-RC1 ]] ; then
 		echo "TODO: bump release"
 	fi
-	obal update -e version=${FULLVERSION%%-RC*} -e prerelease=${FULLVERSION##*-} -e release=keep $RPM_PACKAGES
+	# katello needs dot, lowercase, foreman needs uppercase
+	PRERELEASE=${FULLVERSION##*-}
+	if [[ $PROJECT == katello ]]; then
+		PRERELEASE=".${PRERELEASE,,}"
```
ekohl: What's the `,,` there?

bump_rpm_packaging
```diff
@@ -13,9 +13,14 @@ if [[ $FULLVERSION == *-RC* ]] ; then
 	if [[ FULLVERSION == *-RC1 ]] ; then
 		echo "TODO: bump release"
```
ekohl: For 1.25 we can also move to `rc1` instead of `RC1` to simplify things. The alternative is to implement `release=increment` in obal.

bump_rpm_packaging
```diff
@@ -13,9 +13,14 @@ if [[ $FULLVERSION == *-RC* ]] ; then
...
 	fi
-	obal update -e version=${FULLVERSION%%-RC*} -e prerelease=${FULLVERSION##*-} -e release=keep $RPM_PACKAGES
+	# katello needs dot, lowercase, foreman needs uppercase
+	PRERELEASE=${FULLVERSION##*-}
+	if [[ $PROJECT == katello ]]; then
+		PRERELEASE=".${PRERELEASE,,}"
+	fi
+	obal update --version ${FULLVERSION%%-RC*} --prerelease ${PRERELEASE} -e release=keep $RPM_PACKAGES
```
ekohl:
```suggestion
	obal update --version ${FULLVERSION%%-RC*} --prerelease ${PRERELEASE} --release keep $RPM_PACKAGES
```

bump_rpm_packaging
evgeni: lowercase, same as `| tr [A-Z] [a-z]`

bump_rpm_packaging
evgeni: I'd prefer moving to `rc1`, less difference between katello and foreman, and `release=increment` isn't that simple (if you don't want to add a changelog just yet)

bump_rpm_packaging
ekohl: TIL

bump_rpm_packaging
ekohl: I agree. Debian also uses lowercase and if Katello RPMs also use lowercase, moving Foreman to lowercase would unify them all.

PR #35, opened 2020-02-19
more 2.0 changes

evgeni:
* client bits for 2.0
* proper lowercase rc versioning handling


bump_rpm_packaging
```diff
@@ -9,16 +9,16 @@ cd "$PACKAGING_DIR"
 git fetch $PACKAGING_GIT_REMOTE
 git checkout -b rpm/release-$PROJECT-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
 
-if [[ $FULLVERSION == *-RC* ]] ; then
-	if [[ FULLVERSION == *-RC1 ]] ; then
+if [[ $FULLVERSION == *-RC* ]] || [[ $FULLVERSION == *-rc* ]]; then
+	if [[ $FULLVERSION == *-RC1 ]] || [[ $FULLVERSION == *-rc1 ]]; then
```
ekohl: Bumping the release isn't needed when it's `rc1` (but is with `RC1`). Given we're switching to lowercase and don't expect any more uppercase, I think it's safe to drop this block

releases/client/2.0/settings
```diff
@@ -0,0 +1,3 @@
+FULLVERSION="2.0-rc1"
```
ekohl: Does the client care about the exact version or can we simply say `FULLVERSION=$FOREMAN_VERSION`?

releases/client/2.0/settings
```diff
@@ -0,0 +1,3 @@
...
+FOREMAN_VERSION="2.0"
+OSES="rhel5 rhel6 rhel7 el8 sles11 sles12 fedora29"
```
ekohl: We should prioritize updating to Fedora 30/31 in nightly since F29 is EOL. It feels weird to create a new release on an EOL distro.

releases/client/2.0/settings
evgeni: no, it does not care, you're right.

releases/client/2.0/settings
evgeni: agreed, but not the scope of this pr ;)

bump_deb_packaging
```diff
@@ -9,7 +9,8 @@ cd "$PACKAGING_DIR"
 git fetch $PACKAGING_GIT_REMOTE
 git checkout -b deb/release-$PROJECT-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
 
-scripts/changelog.rb -v ${FULLVERSION/-RC/~rc}-1 -m "$FULLVERSION released" debian/*/*/changelog
+DEBIANIZED_VERSION="$(echo ${FULLVERSION} | sed 's/-RC/~rc/i')"
```
ekohl: Would it be safe to assume we only have lowercase rcs now?

evgeni: you mean, we're not going to do any pre 2.0 RCs anymore? maybe.

ekohl: Exactly. We've never done RCs for GA versions (like a 1.24.2-RC1)

bump_rpm_packaging
```diff
@@ -9,16 +9,13 @@ cd "$PACKAGING_DIR"
 git fetch $PACKAGING_GIT_REMOTE
 git checkout -b rpm/release-$PROJECT-$FULLVERSION $PACKAGING_GIT_REMOTE/$REMOTE_BRANCH
 
-if [[ $FULLVERSION == *-RC* ]] ; then
-	if [[ FULLVERSION == *-RC1 ]] ; then
-		echo "TODO: bump release"
-	fi
+if [[ $FULLVERSION == *-RC* ]] || [[ $FULLVERSION == *-rc* ]]; then
 	# katello needs dot, lowercase, foreman needs uppercase
```
ekohl: lowercase/uppercase also looks outdated now

PR #36, opened 2020-02-21
drop rails tag from releases newer than 1.24

evgeni:


evgeni:
We identically disabled rails related things for 1.22 and older in Jenkins, so this is just consistently broken ;)

PR #45, opened 2020-05-12
Add Foreman 2.1 client settings

ehelms:


releases/client/2.1/settings
```diff
@@ -0,0 +1,3 @@
+FOREMAN_VERSION="2.1"
+FULLVERSION=$FOREMAN_VERSION
+OSES="rhel5 rhel6 rhel7 el8 sles11 sles12 fedora29"
```
ekohl, line 3: Fedora 29 is EOL (and so is 30). Should we just not publish it?

ehelms: I think we need to ask that question prior to branching. I'd rather we take what we have been testing and building via nightly and release it. But as a fix to this looking ahead, we should plan out the client supported OSes as part of the release planning thread that @tbrisker starts each release.

ekohl: The fact we've built Fedora 29 in nightly doesn't mean we need to publish it. My suggestion wasn't to start building F31 and F32 for 2.1 but rather not deliver Fedora client packages at all.

PR #51, opened 2020-07-14
Katello 3.16 config

pcreech:


releases/katello/3.16/settings
```diff
@@ -0,0 +1,4 @@
+FULLVERSION='3.16.0-rc4'
```
ekohl, line 1: I think by now we're at `3.16.0`

ehelms: We're actually at rc4.1

pcreech: Technically, it's 3.16.0-rc4.1 iirc, but this is adding my work from last week,to be built upon by the person doing the next release.

pcreech: I'll defer to @ehelms if he wants me to update to rc4.1 here or if he'll handle it after this merge

releases/katello/3.16/settings
```diff
@@ -0,0 +1,4 @@
...
+FOREMAN_VERSION='2.1'
+OSES="rhel7 el8"
```
ehelms, line 3: I think this should be only rhel7 now

pcreech: I want to say i tried that, but still got errors on release pipelines.  I'll remove, shoudln't be hard to put back if it dies.

PR #54, opened 2020-08-04
Update GPG export message to point at FIXME

ehelms:


export_gpg_public
```diff
@@ -17,7 +17,7 @@ CHECKOUT="${GIT_DIR}/theforeman.org"
 
 gpg2 --homedir "$KEYDIR" --armor --export "$FULLGPGKEY" > "${CHECKOUT}/static/keys/${FULLGPGKEY}.pub"
 
-echo "Add this to ${CHECKOUT}/security.md"
+echo "Fill in any FIXME's and then add this to ${CHECKOUT}/security.md"
 gpg2 --homedir "$KEYDIR" --fingerprint "$FULLGPGKEY"
```
ekohl, line 21: This line should print all the info you need (and I didn't want to parse that output). Is a hint for that needed as well or is that clear enough?

ehelms: That's clear enough when looking at the screen. The problem is more that I read "Add this to" and just added it like it told me to.

PR #56, opened 2020-08-12
Add an all-in-one foreman sign script

ehelms:


release_foreman_rpms
```diff
@@ -0,0 +1,18 @@
...
+VERSION=$1 ./sign_rpms
+VERSION=$1 ./upload_rpm_signatures
+VERSION=$1 ./upload_rpms
+VERSION=$1 ./process_rpms
+
+VERSION=$1 PROJECT=client ./download_rpms
+VERSION=$1 PROJECT=client ./sign_rpms
+VERSION=$1 PROJECT=client ./upload_rpm_signatures
+VERSION=$1 PROJECT=client ./upload_rpms
+VERSION=$1 PROJECT=client ./process_rpms
```
ekohl, line 18: I'd prefer not to repeat this. That way you can use the same script for all projects (foreman, client, katello). It could even call itself:
```suggestion
if [[ $PROJECT == foreman ]] ; then
    PROJECT=client $0 $1
fi
```

ehelms: Done... with a comment. I prefer being explicit but given this was your original project I've opted for DRY with a comment.

release_foreman_rpms
```diff
@@ -0,0 +1,17 @@
...
+  exit 1
+fi
+
+VERSION=$1 ./download_rpms
+VERSION=$1 ./sign_rpms
+VERSION=$1 ./upload_rpm_signatures
+VERSION=$1 ./upload_rpms
+VERSION=$1 ./process_rpms
+
+# Run the same steps for client but avoid recursion
```
ekohl, line 14: Well, you can now also call it for Katello and that doesn't imply a client release. You can also run it just for a client release.

release_foreman_rpms
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+
+if [ -z "$1" ];then
+  echo "Pass version to release, e.g. ./release_foreman_rpms 2.2"
+  exit 1
+fi
+
+VERSION=$1 ./download_rpms
+VERSION=$1 ./sign_rpms
```
ekohl, line 9: My issue here with blindly signing is that for every release I did I always checked if the downloaded RPMs were ones that I actually expected. If you're just going to sign without any checks, what's the signing really worth?

That was the idea behind downloading the RPMs to a directory: you can inspect them there. Then you perform the signing operations. Once done, they're moved so you have an archive.

ehelms: I'm not sure that idea has been captured anywhere if I look at the README or our Foreman procedures. If I am someone following either of those, I would just go step by step with no inspection. And, if I am being honest, I don't know what I would be inspecting the RPMs **for**.

ekohl: I always look at the version numbers and if they make sense. For example, is some change expected? Some nightly version of hammer should not be signed and uploaded.

ehelms:
I pushed a new update with new scripts that I think more capture what I am aiming for here. Which is to group commonly run tasks together and derease time to release by having to remember to come back to my terminal and run the next item.

release_foreman_rpms
ehelms: That feels like something we should codify through documentation in our procedure, with the list of things to look for, and where possible code to validate the rules otherwise I don't think we can expect consistency among releasers.

ekohl (changes requested):
I still think downloading and signing should be separate steps.

PR #57, opened 2020-09-01
add pulpcore signing settings

evgeni:


releases/pulpcore/settings
```diff
@@ -0,0 +1,5 @@
+OSES="el7 el8"
+KOJI_TAGS="pulpcore-@VERSION@-@OS@"
+PASS_NAME_GPG="theforeman/releases/pulpcore/gpg-passphrase"
+PASS_NAME_KEY="theforeman/releases/pulpcore/gpg-key"
+FULLGPGKEY=""
```
evgeni, line 5: this is intentional, as otherwise it wants to use the foreman key, but I don't want to be tight to foreman releases.

releases/pulpcore/3.6/settings
```diff
@@ -0,0 +1,3 @@
+FOREMAN_VERSION='2.1'
```
evgeni, line 1: if I don't set that, the scripts bail out, and adding "if not pulpcore" seemed also weird?

evgeni:
doesn't strictly belong in this PR, but it's as good place to discuss as any other.

I've created the key with this PR, but dropped the "VERSION" comment to the key, as I'd prefer not to generate new keys per release (this is also what Pulp 2 has been doing).

@ehelms @ekohl @pcreech @zjhuntin  what do you think?

releases/pulpcore/3.6/settings
ekohl: Do we need to inherit anything? Perhaps we can support `FOREMAN_VERSION=''` or `FOREMAN_VERSION='none'`.

evgeni: I've implemented "none" now, that was a good idea.

ekohl (commented):
I wanted to verify the GPG key so I ran:

```
PROJECT=pulpcore VERSION=3.6 ./import_gpg_private
```

This didn't work for me.

We have https://yum.theforeman.org/pulpcore/3.4/ as well so a 3.4 config might be nice to have.

If we compare it to https://yum.theforeman.org/releases/2.2/ we also have a RPM-GPG-KEY-foreman file. Probably good to to a similar thing for pulpcore. That makes it easy to consume since Forklift can just point there for the GPG key.

evgeni:
@ekohl well, I forgot to export the key to gopass :see_no_evil: 

How it that key deployed to yum.tfm? I'd gladly take that, and then also use a versioned key (a nice distribution way was why I didn't want versioned keys)

ekohl:
> How it that key deployed to yum.tfm? I'd gladly take that, and then also use a versioned key (a nice distribution way was why I didn't want versioned keys)

Currently manually but we can automate it. The export script already knows how to get the GPG key and you have pretty much all the variables (`HOSTS` in settings + versions). You may need to update `DOCROOT` to `DOWNLOADS_DOCROOT` and introduce `YUM_DOCROOT`. `sign_tarballs` can be a good inspiration.

releases/pulpcore/3.6/settings
```diff
@@ -0,0 +1,3 @@
+FOREMAN_VERSION='none'
```
ekohl, line 1: This can live in `releases/pulpcore/settings` since that's loaded prior to foreman.

evgeni: Nope:
https://github.com/theforeman/theforeman-rel-eng/blob/d8af5d364b2ab5c5040489eecca8e5d181a252e1/settings#L13-L17

We'd have to change that loading logic, which I tried to avoid for now.

ekohl: Ah, line 13 is RELEASEDIR and not PROJECTDIR. It should be safe to load that as well, right? We load it around line 50 as well.

evgeni: Yeah, I made a `load_settings` function now, how is that?

ekohl (approved):
https://github.com/theforeman/theforeman-rel-eng/blob/master/export_gpg_public doesn't entirely work anymore. It should use `$SIGNER_NAME` instead of hardcoding the name (or extract it from the key).

What does make sense, is an easy script to verify the key. For example, in this PR I saw the Pulp 3.6 settings and wanted to verify the key. The best way that I know of:
```
PROJECT=pulpcore VERSION=3.6 ./import_gpg_private
PROJECT=pulpcore VERSION=3.6 ./export_gpg_public
```

Ideally this would be a section in README.

Using the above method I have verified the listed keys are available and can reproduce the signature. The code also looks good so I think this can be merged.

evgeni:
*technically* the `export_gpg_public` had a wrong key name even before, but I'll fix that.

PR #60, opened 2020-09-16
upload tarball signatures as the downloads user, not root

evgeni:


evgeni:
this is so that https://github.com/theforeman/foreman-infra/pull/1451 don't get broken.

ekohl (approved):

PR #63, opened 2020-09-25
Pulpcore 3.7

evgeni:


import_gpg_private
```diff
@@ -7,4 +7,4 @@ if [[ ! -d "$KEYDIR" ]] ; then
 fi
 
 copy_gpg_key
-gopass show "$PASS_NAME_KEY" | gpg2 --homedir "$KEYDIR" --import
+gopass show --password "$PASS_NAME_KEY" | gpg2 --homedir "$KEYDIR" --import
```
ehelms, line 10: Does this change mean everyone needs newer gopass clients?

generate_gpg
```diff
@@ -15,7 +15,7 @@ fi
 
 mkdir -m 0700 $KEYDIR
 
-( gopass show "$PASS_NAME_GPG" 2> /dev/null || gopass generate "$PASS_NAME_GPG" 20 ) > /dev/null
+( gopass show --password "$PASS_NAME_GPG" 2> /dev/null || gopass generate "$PASS_NAME_GPG" 20 ) > /dev/null
```
ekohl, line 18: Do we need a wrapper to get a password?

import_gpg_private
evgeni: No, this feature has been there for a long time: https://github.com/gopasspw/gopass/commit/8fa13d84e3656cfc4ee6717f5f485c9e471ad996

It's just the output of `show` without `--password` changed recently.

generate_gpg
evgeni: Could… but really need? Nah.

import_gpg_private
evgeni: Adding more detail.

This is a change in gopass 1.10 (https://github.com/gopasspw/gopass/blob/master/CHANGELOG.md#1100) how it stores secrets. Before it would just store them "plain", so if your secret was "secret", it stored "secret", now it stores "Password: secret" to allow more than a single entry to be present.

ekohl (approved):

PR #75, opened 2020-11-17
Invalidate CDN when uploading signatures

evgeni:


sign_tarballs
```diff
@@ -14,9 +14,11 @@ upload() {
 	local files="{'files': {"
 	for project in $TAR_PROJECTS ; do
 		filename="$project-$FULLVERSION.tar.bz2"
-		files+="'${TARDIR}/${filename}.sig': '${DOCROOT}/${project}/${filename}.sig',"
+		files+="'${TARDIR}/${filename}.sig': '${project}/${filename}.sig',"
 	done
-	files+="}}"
+	files+="},"
+	files+="{'docroot': '${DOCROOT}'}"
+	files+="}"
```
ekohl, line 21: And my monstrosity keeps growing ... At some point it may be needed to write a helper to generate JSON via some tool like Python.

ekohl (approved):

PR #84, opened 2021-02-01
add foreman 2.4 settings

evgeni:


ekohl (approved):
I submitted https://github.com/theforeman/theforeman-rel-eng/pull/85 before I saw this one. GPG key is correct.

PR #86, opened 2021-02-01
add 2.4 client config

evgeni:


releases/client/2.4/settings
```diff
@@ -0,0 +1,3 @@
+FOREMAN_VERSION="2.4"
+FULLVERSION=$FOREMAN_VERSION
+OSES="rhel6 rhel7 el8 sles11 sles12"
```
ekohl, line 3: Will we keep building for RHEL6? CentOS 6 is EOL now.

evgeni: Given our downstream still supports it, I'd prefer to keep it.

PR #91, opened 2021-03-19
Various settings files updates

pcreech:
client 2.3 and 2.4 FULLVERSION fixes
Forean 2.4.0-rc3
Katello 3.18.1
Katello 3.17.3


releases/client/2.3/settings
```diff
@@ -1,3 +1,3 @@
 FOREMAN_VERSION="2.3"
-FULLVERSION=$FOREMAN_VERSION
+FULLVERSION="2.3.0-RC1"
```
ekohl, line 2: Why? This was fine as it was IMHO. It's not actually used anyway and better than something which isn't true either.

releases/client/2.4/settings
```diff
@@ -1,3 +1,3 @@
 FOREMAN_VERSION="2.4"
-FULLVERSION=$FOREMAN_VERSION
+FULLVERSION=2.4.0-rc3
```
ekohl, line 2: Same as for 2.3: I think this was correct.

releases/client/2.3/settings
pcreech: There is a step in the process where it expects the full version, not just the x.y, and it dies every time.

evgeni: Not for the client, no

pcreech: Then.... why does it fail and I have to make the edit. every. single. time.

evgeni: *what* does fail?

evgeni: Ah, release_pipeline script for client fails because ci-jobs expects the full version... Mhhh

pcreech: There ya go!  I was going through the rpm scripts but looks like you got to it.

ekohl: I still don't really see how that would fail. For the client pipeline a version with major.minor should suffice and we have that here.

releases/foreman/2.4/settings
```diff
@@ -1,2 +1,2 @@
-FULLVERSION='2.4.0-rc2'
+FULLVERSION='2.4.0-rc3'
```
ekohl, line 1: It's been updated to GA now so there's a conflict.

releases/client/2.3/settings
evgeni: https://github.com/theforeman/theforeman-rel-eng/blob/622870650969aa3367e298d532db2f6230623b8b/ci-jobs#L130-L133

ekohl: Ah right. Instead of changing these settings files with confusing out of date info, I think we should instead change the parsing to handle major.minor.

ekohl: I prefer to merge https://github.com/theforeman/theforeman-rel-eng/pull/92 instead.

ekohl (changes requested):
Please rebase and update.

PR #99, opened 2021-05-18
Update katello version to 4.1-rc2

zjhuntin:


evgeni (approved):
Technically not required as nothing checks/uses fullversion of Katello

PR #104, opened 2021-06-09
update config for Foreman 2.3.5

evgeni:


releases/foreman/2.3/settings
```diff
@@ -1,3 +1,2 @@
-FULLVERSION='2.3.4'
+FULLVERSION='2.3.5'
 FULLGPGKEY='1CC363863DD64AF30638DB79C101586BE0745924'
-OSES="rhel7"
```
ekohl, line 3: Is this no longer relevant or was it wrong to begin with?

evgeni: it was wrong to begin with
 2.3 works just fine on el8, and that's the default in the global  `settings` for a long time now.

ekohl (approved):

releases/foreman/2.3/settings
evgeni: added here: https://github.com/theforeman/theforeman-rel-eng/commit/75d8fa9f58784e2ca7f334463b23d4077d1d8a9

no idea why

PR #122, opened 2021-11-09
GPG Keys for Foreman 3.1

Odilhao:


releases/client/3.1/settings
```diff
@@ -0,0 +1,3 @@
+FOREMAN_VERSION="3.1"
+FULLVERSION=$FOREMAN_VERSION
+OSES="rhel6 rhel7 el8 sles11 sles12"
```
ekohl, line 3: We normally always have line endings in files.

ekohl (changes requested):
I've confirmed I can import the GPG key from our shared location. Please fix the line ending.

releases/client/3.1/settings
Odilhao: Done, I was not aware of that.

ekohl (approved):

PR #136, opened 2022-02-22
Add a script to control Redmine

ekohl:
For now this is limited to versions.

The close_redmine_version script which does the right thing during the release process.

There is no create_redmine_version script since that requires some calculation of the next version and I didn't want to touch that just yet. The direct redmine script can be used instead.

### DIFF
--- a/close_redmine_version
+++ b/close_redmine_version
@@ -1,0 +1,7 @@
+#!/bin/bash -e
+
+. settings
+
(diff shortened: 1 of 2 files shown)


zjhuntin:
Hey, what would be a good way to test this?

ehelms (approved):
This looks sane enough code wise, let's take it and iterate if we find bugs

PR #153, opened 2022-06-08
don't mess around with ~/.rpmmacros

evgeni:
users might want to have own macros defined there, so let's just set
them on the CLI.


evgeni:
cc @Odilhao this should unfuck the thing you've been seeing with your python path :)

PR #171, opened 2022-10-19
Use new secret storage

ekohl:
puppetmaster.theforeman.org is being decommissioned. A new hostname is introduced to provide the service, rather than the actual server name.

Currently a draft since the secrets have not been migrated yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,23 +20,11 @@ Ensure that `gopass` is initialized after installing the first time (and that yo
 gopass init <YOUR-PUB-KEY-HASH>
 ```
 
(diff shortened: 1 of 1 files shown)


evgeni:
Which reminds me that once we do the switch, we need to tell existing users how to flip their existing installs.

Or should they wipe and re-setup?

ekohl:
Users can show the locations with `gopass mounts` and then use regular `git remote`. That's probably easiest. The alternative is `gopass mounts rm` and then re-add.

ekohl:
This is now live. See https://github.com/theforeman/foreman-infra/pull/1809#issuecomment-1385755155 as well.

PR #180, opened 2022-12-16
Set OSES to el8 on Foreman 3.4+

ekohl:
This sets the default to el8 and sets it explicitly on < 3.4 for historical context.


releases/foreman/settings
```diff
@@ -2,3 +2,4 @@ TAR_PROJECTS="foreman foreman-proxy foreman-installer foreman-selinux"
 RPM_PACKAGES="foreman foreman-installer foreman-proxy foreman-release foreman-selinux"
 KOJI_TAGS="foreman-@VERSION@-@OS@"
 PACKAGING_SUBDIR="packages/foreman"
+OSES="el8"
```
ekohl, line 5: I was hesitant to change it in `settings`. Perhaps we should drop it altogether there and enforce all projects to set it explicitly?

evgeni: sounds fair to me

Issue #193, opened 2023-03-13
Add script to setup a local release environment with each repository required for tagging


ehelms:
Currently the scripts assume a particular location and setup of the needed repositories when releasing. This can be overwritten with a variable, but still requires all projects to be co-located. If we provide a script to setup the environment and have the releaser use it instead there is less confusion and more consistency.

ehelms:
Another reason is that `tag_push` assumes a branch name of `upstream` to be able to push to.

ekohl:
Yes, it is a bit tuned to my setup. I opened https://github.com/theforeman/theforeman-rel-eng/pull/196 to address this and other issues.

ehelms:
Thanks for #196, I think that helps a lot. I'd like to still keep this idea open as having a script to set up an environment instead of having to go figure it out manually especially someone new to the process will help a lot.

ekohl:
Thinking more about it, it does fit in the general pattern to have (very) small and dedicated scripts. An `ensure_clean_git_checkouts` script that only does the cloning or yell about dirty work spaces makes sense to me.

PR #195, opened 2023-03-13
Set TAR_PROJECTS for foreman to use smart-proxy

ehelms:


ehelms:
Hrmm, I guess this breaks downloading the tarballs, so maybe a different variable is needed or https://github.com/theforeman/theforeman-rel-eng/issues/193 could address that by normalizing the environment.

ekohl:
In https://github.com/theforeman/theforeman-rel-eng/pull/196 I use an alternative way that doesn't break tarballs.

ehelms:
Closing in favor of #196

PR #198, opened 2023-03-13
Set ansible_become_password automatically

### LINKED ISSUE #197: Set Ansible become password via parameters
Right now we use an approach where the password is put in the paste buffer and then asked on the command line. It is possible to set an `ansible_become_password` parameter (https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_privilege_escalation.html#become-connection-variables) to automate this.

ekohl:
This avoids the need to abuse the paste buffer and requires fewer manual steps.

Currently untested

Fixes #197

### DIFF
--- a/settings
+++ b/settings
@@ -73,8 +73,19 @@ copy_gpg_key() {
 	gopass show -c "$PASS_NAME_GPG"
 }
 
-copy_unix_pass() {
-	gopass show -c "theforeman/unix"
+# Invoke ansible-playbook with ansible_become_password set safely
+ansible_playbook() {
+        local password_file password
+        password_file=$(mktemp --suffix=.json)
+        chmod 0600 "$password_file"
+
+	password=$(gopass show "theforeman/unix" | head -n 1)
+	cat > "${password_file}" <<-ANSIBLE_VARS
+        {"ansible_become_password": "${password}"}
+	ANSIBLE_VARS
+
+	ansible-playbook --extra-vars="@${password_file}" "$@"
+	rm -f "${password_file}"
 }
 
 # vim: ft=sh
--- a/sign_tarballs
+++ b/sign_tarballs
@@ -20,8 +20,7 @@ upload() {
 	files+="'docroot': '${DOCROOT}'"
 	files+="}"
 
-	copy_unix_pass
-	ansible-playbook --ask-become-pass --inventory=$HOSTS --extra-vars="$files" upload_tarball_signatures.yaml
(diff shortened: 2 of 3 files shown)


sign_rpms
```diff
@@ -2,17 +2,20 @@
 
 . settings
 
-copy_gpg_key
+gpg_pass=$(mktemp)
+trap 'shred --remove "${gpg_pass}"' EXIT
+chmod 0700 "$gpg_pass"
+show_gpg_password > "$gpg_pass"
+
+SIGN_FLAGS="--define '_gpg_sign_cmd_extra_args --pinentry-mode loopback --passphrase-file ${gpg_pass}' --define '_gpg_path $KEYDIR' --define '_gpg_name $SIGNER'"
```
ekohl, line 10: TODO: this should probably become an array

settings
```diff
@@ -69,12 +69,27 @@ load_settings
...
+
+# Invoke ansible-playbook with ansible_become_password set safely
+ansible_playbook() {
+        local password_file password
+        password_file=$(mktemp --suffix=.json)
+        chmod 0600 "$password_file"
+
+	password=$(gopass show "theforeman/unix" | head -n 1)
+	cat > "${password_file}" <<-ANSIBLE_VARS
+        {"ansible_become_password": "${password}"}
```
ekohl, line 88: This assumes `"` is not in the password.

ekohl (commented):
@evgeni while this is still work in progress, I'd appreciate your thoughts if I'm doing anything weird.

sign_rpms
pcreech: What are your thoughts on inserting TODO items as comments in the source files in the case of these scripts?  I think this would help identify areas that need improvement in reviewing the code.

pcreech (approved):
In concept, everything looks straightforward and it appears to handle one of my bigger issues with signing rpms.  (the need to ctl-v the password for signing).

I'm in favor of merging, and following up with subsequent PRs if we identify any major issues.  

My only question is how do we capture the TODO mentioned on the PR to be addressed at a later time? (the turning the SIGN_FLAGS defines into an array instead of string)

sign_rpms
ekohl: I think that we should apply shellcheck on this repository, which highlighted it locally for me. I'll take a quick look at fixing it.

pcreech:
Nice.  ACK on the updates.  Merging

PR #201, opened 2023-03-15
Extend branching and releasing scripts

### LINKED ISSUE #193: Add script to setup a local release environment with each repository required for tagging

ekohl:
These contain various improvements. See individual commits for details.

Fixes #193

### DIFF
--- a/branch_project
+++ b/branch_project
@@ -1,0 +1,19 @@
+#!/bin/bash -e
+
+. settings
+
+for project in $(./ensure_clean_git_checkouts "$@") ; do
+	(
+		cd "$GIT_DIR/$project"
+		git fetch --quiet "$GIT_REMOTE"
+		git checkout -b "${VERSION}-stable" "${GIT_REMOTE}/develop"
+
+		if [[ $project == foreman-installer ]]; then
+			bundle exec rake pin_modules
+			sed -i '/Puppetfile.lock/d' .gitignore
+			bundle exec librarian-puppet install
+			git add .gitignore Puppetfile*
+			git commit --message "Pin Puppet modules for ${VERSION}"
+		fi
+	)
+done
--- a/branch_push
+++ b/branch_push
@@ -1,0 +1,10 @@
+#!/bin/bash -e
+
+. settings
+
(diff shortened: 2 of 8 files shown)


branch_project
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -xe
```
ekohl, line 1: Note to self: I should update this to just `-e`. `-xe` is the debug version.

Having said that, @ehelms raised that it can be difficult to figure out what's going on. Would it make sense to have a `DEBUG` env var that calls `set -x`?

branch_project
```diff
@@ -0,0 +1,29 @@
...
+
+. settings
+
+if [[ -z $1 ]] ; then
+	PROJECTS=$TAR_PROJECTS
+else
+	PROJECTS=$*
+fi
+
+PROJECTS=${PROJECTS/foreman-proxy/smart-proxy}
```
ehelms, line 11: Lines 5-11 are repeated fairly often across a few scripts, is there a way to capture them into a re-usable piece?

ekohl: I thought about the same, but struggled a bit. I'm now thinking about enhancing the ensure script to also output the project names and then use those as input. Will try something.

ekohl: How about this version?

ehelms (approved):
Looks sane and clean, let's give it a go

develop_branch_bump
ekohl, line 1: Manually tested. While testing I realized I forgot to commit the changes, which is now added.

branch_push
ekohl, line 1: I didn't test this, but it's simple enough that it's easy to fix on the fly if needed.

develop_branch_push
ekohl, line 1: I didn't test this, but it's simple enough that it's easy to fix on the fly if needed.

ekohl (commented):
Only changes since your review are that `develop_branch_bump` now creates a git commit.

PR #173, opened 2022-11-07
Add a script to transform the nightly tool_belt config

ekohl:
This takes the nightly config as an input and transforms it into a release version. There are still some TODOs, such as adding inputs and adding the actual release.

PR #202, opened 2023-03-15
Add a script to show the GPG key to be used for Koji

ekohl:
This value was already calculated to download from Koji. Having a script to print it simplifies writing the mashing scripts.


ekohl:
I'm also tempted to push for https://github.com/theforeman/theforeman-rel-eng/pull/173 instead. I was much further with that than I thought.

PR #214, opened 2023-04-04
Add Katello compatibility to tool_belt config generation

ekohl:
Includes https://github.com/theforeman/theforeman-rel-eng/pull/213 and is my current work in progress. It still contains issues, which are marked as TODO.

### DIFF
--- a/build_tool_belt_config
+++ b/build_tool_belt_config
@@ -2,13 +2,14 @@
 
 import argparse
 import sys
+from typing import Union, Tuple
 
 from redminelib import Redmine
 from ruamel.yaml import YAML
 
 REDMINE_URL = 'https://projects.theforeman.org'
 
-def find_redmine_versions(project_id: str, prefix: str) -> dict[str, int]:
+def find_redmine_versions(project_id: str, prefix: Union[str, Tuple[str, ...]]) -> dict[str, int]:
     """
     Find all X.Y.Z versions in Redmine based on version X.Y
     """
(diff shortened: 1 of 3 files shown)


build_tool_belt_config
```diff
@@ -26,16 +27,25 @@ def create_mash_config(version: str, gpg_key: str, redmine_project: str) -> None
     yaml.indent(mapping=2, sequence=4, offset=2)
     yaml.explicit_start = True
 
+    # TODO: foreman should NOT be replaced
```
ekohl, line 30: @wbclark @ianballou I didn't get around to this today, but it ends up creating `foreman-4.9-el8` and `foreman-plugins-4.9-el8` in the config which should be `3.7`.

I tried this:
```python
    data = yaml.load(
        re.sub(f'(?P<tag>{redmine_project}(-[a-z]+)?)-nightly', f'\\1-{version}', sys.stdin.read())
    )
```

But then figured out I also needed to pass in the Foreman version, which is going to be more work.

For now I'd recommend using this script and manually replace the instances. It's just a few anyway.

create_mash_configs
```diff
@@ -9,7 +9,11 @@ CONFIG_DIR="${TOOL_BELT}/configs/${PROJECT}"
 OUTPUT_FILE="${CONFIG_DIR}/${VERSION}.yaml"
 OUTPUT_DIR="${TOOL_BELT}/mash_scripts/${PROJECT}/${VERSION}.0"
 
-./build_tool_belt_config "$VERSION" "$GPGKEY" "$PROJECT" < "${CONFIG_DIR}/nightly.yaml" > "${OUTPUT_FILE}"
+if [[ $PROJECT == katello ]] ; then
```
ekohl, line 12: This is not pretty, but it works.

ekohl:
Foreman has stopped using tool_belt and I wonder what Katello will do with it. It's now just used for cherry picking, but I don't intend to finish this.

PR #224, opened 2023-05-24
Make branch_* Katello compatible

ekohl:
Katello follows a different branch naming convention. It uses master instead of develop and prefixes stable branches with KATELLO- instead of a -stable suffix.

<del>Right now it at least doesn't have the correct projects set up in `ensure_clean_git_checkouts`. We don't set `TAR_PROJECTS` and it assumes everything lives in the `theforeman` namespace.</del>

@lfu or @chris1984 are there more steps that you take when branching?

cc @wbclark


chris1984:
I can look Monday, sorry for missing this one

chris1984 (approved):
Looks good to me, can't really think of what else to add offhand

ekohl:
@chris1984 I've added a commit that now uses it in the procedure. Mind having another look?

chris1984 (approved):
Looks good, only suggestion I would make is to take this out of the readme since you have moved the procedures here.

`Procedures
Procedures are kept in [tool_belt](https://github.com/theforeman/tool_belt) with small wrappers in this repository.`

procedures/katello/branch.md.erb
```diff
@@ -64,10 +64,10 @@
 
 ## Release Owner
 
-- [ ] Create KATELLO-<%= release %> branches
-  - [ ] [katello](https://github.com/katello/katello)
-- [ ] Bump versions to <%= develop %>-master
-  - [ ] [katello](https://github.com/katello/katello)
+- [ ] Create KATELLO-<%= release %> branches using [branch_project](https://github.com/theforeman/theforeman-rel-eng/blob/master/branch_project)
```
chris1984, line 67: I would change `branches` to `branch`

procedures/katello/branch.md.erb
```diff
@@ -64,10 +64,10 @@
...
+- [ ] Push KATELLO-<%= release %> branches using [branch_push](https://github.com/theforeman/theforeman-rel-eng/blob/master/branch_push)
```
chris1984, line 68: I would change `branches` to `branch`

procedures/katello/branch.md.erb
ekohl: Oh yes, for Katello it's one branch. In Foreman the same script does it for multiple branches.

ekohl:
> Looks good, only suggestion I would make is to take this out of the readme since you have moved the procedures here.

Good catch.

PR #234, opened 2023-06-23
Add Pulpcore 3.28 GPG keys

Odilhao:


releases/pulpcore/3.28/settings
```diff
@@ -0,0 +1,3 @@
+FULLVERSION='3.28.0'
+FULLGPGKEY='C9A0A8F7C3B595516AE0247387EDE3EC2622DB63'
+OSES="el8 el9"
```
ekohl, line 3: Can you ensure a newline?

ekohl (commented):
I've confirmed I can import the private key with the same public key.

releases/pulpcore/3.28/settings
Odilhao: Oh boy here we go again :smile:

ekohl (approved):

PR #244, opened 2023-08-08
Procedure: switch to katello release branch before pushing tags

ianballou:
Re-opening of https://github.com/theforeman/tool_belt/pull/552

As for the comments about having these steps automated -- I think that's a good next step, but in the meantime, I'd like to get the procedure here corrected so that releasers do the tag pushing correctly.


ekohl (approved):
I agree this a good intermediate step.

PR #247, opened 2023-08-09
Add a rel_eng_script macro in procedures

ekohl:
This makes it easier to link to theforeman-rel-eng scripts.

Right now this only applied to a single procedure, but it gives a good impression of the changes.


evgeni (approved):

evgeni:
You want to merge this as-is, or apply the change to all procedures?

ekohl:
> You want to merge this as-is, or apply the change to all procedures?

I at least did Foreman & Katello. I don't see the Pulp procedures posted so I didn't touch those. Also took the chance to use more scripts that already existed (see last commit).

evgeni (approved):

PR #207, opened 2023-03-28
Support git worktrees for foreman-packaging

ekohl:
Previously it was assumed to have something like `~/dev/foreman-packaging` but with git worktrees you can create multiple working directories. I have switched over to:

```console
$ tree -L 1 ~/dev/foreman-packaging
/home/ekohl/dev/foreman-packaging
├── deb-3.1
├── deb-3.2
├── deb-3.3
├── deb-3.4
├── deb-3.5
├── deb-3.6
├── deb-develop
├── master
├── rpm-3.1
├── rpm-3.2
├── rpm-3.3
├── rpm-3.5
├── rpm-3.6
└── rpm-develop
```

This is all backed by a single .git directory, which makes it much cheaper.

It is off by default, but can be enabled with a (local) setting.

PR #254, opened 2023-08-23
Don't use obal to git branch & commit + unify bump packages

ekohl:
Obal uses a complex branch name and an obscure commit message. This simplifies the whole workflow and brings it more in line with the Debian side. After bringing those closer together, the next script introduces a common `bump_packaging` script. The old scripts remain as very light wrappers. The last commit is only there to make shellcheck happy. In practice it wouldn't have been any problem since package names never have spaces, but it's still a good practice.


ekohl:
I used this (together with #207) for foreman-3.8.0-rc1: https://github.com/theforeman/foreman-packaging/pull/9766 & https://github.com/theforeman/foreman-packaging/pull/9767

ekohl:
Now updated and I tested it with RPMs where it previously didn't add the files that obal changed.

evgeni (approved):

PR #257, opened 2023-08-23
Explict instructions using commands to prepare build systems

ekohl:
This is to keep my notes. I think this could be captured in a script now, given it's just steps to execute. Will work on that tomorrow.


ekohl:
I didn't have time for a script, so I rebased this and marked it ready for review.

PR #262, opened 2023-09-01
Define the reposync URL in line and fetch modulemd remotely

ehelms:
This avoids the need to carry around a reposync config file and will also nicely match the supplied version. The script still needs the modulemd file stored in foreman-packaging so this fetches that remotely from github.

### DIFF
--- a/build_stage_repository
+++ b/build_stage_repository
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 from subprocess import check_output, STDOUT, CalledProcessError
+from urllib.request import urlretrieve
 import os
 import yaml
 import glob
@@ -8,9 +9,10 @@ import shutil
 import time
 import hashlib
 import sys
+import tempfile
 
 
-def move_rpms_from_copr_to_stage(collection, version, src_folder, dest_folder, source=False):
+def move_rpms_from_copr_to_stage(collection, version, src_folder, dest_folder, dist, arch, source=False):
     if source:
         print(f"Moving {collection} Source RPMs from Copr directory to stage repository")
     else:
@@ -19,7 +21,7 @@ def move_rpms_from_copr_to_stage(collection, version, src_folder, dest_folder, s
     if not os.path.exists(dest_folder):
         os.mkdir(dest_folder)
 
-    repo_folder = f"{src_folder}/el8-{collection}-{version}"
+    repo_folder = f"{src_folder}/{dist}-{collection}-{version}-{arch}"
 
     if source:
         files = glob.glob(repo_folder + f"/**/*.src.rpm")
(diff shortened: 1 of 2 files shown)


build_stage_repository
```diff
@@ -34,7 +36,8 @@ def move_rpms_from_copr_to_stage(collection, version, src_folder, dest_folder, s
 
 
 def modulemd_yaml(collection):
-    return f"modulemd/modulemd-{collection}-el8.yaml"
+    path, headers = urlretrieve(f"https://raw.githubusercontent.com/theforeman/foreman-packaging/rpm/develop/modulemd/modulemd-{collection}-el8.yaml")
```
ekohl, line 39: This should take the branch into account:
```suggestion
    path, headers = urlretrieve(f"https://raw.githubusercontent.com/theforeman/foreman-packaging/rpm/{branch}/modulemd/modulemd-{collection}-el8.yaml")
```

build_stage_repository
```diff
@@ -70,7 +73,7 @@ def create_modulemd(collection, version, stage_dir):
     with open(modulemd_yaml(collection), 'r') as file:
         modules = yaml.safe_load(file)
 
-    modules['data']['artifacts'] = {'rpms': output.splitlines()}
+    modules['data']['artifacts'] = {'rpms': [str(rpm, 'UTF-8') for rpm in output.splitlines()]}
```
ekohl, line 76: If you use `universal_newlines=True` when calling `check_output` there is no need to decode (in my experience)

build_stage_repository
```diff
@@ -96,9 +100,9 @@ def sync_copr_repository(collection, version, target_dir, source=False):
         'reposync',
         '--newest-only',
         '--repo',
-        f"el8-{collection}-{version}",
-        '--config',
-        'reposync_config.conf',
+        f"{dist}-{collection}-{version}",
+        '--repofrompath',
+        f"{dist}-{collection}-{version},https://download.copr.fedorainfracloud.org/results/@theforeman/{collection}-{version}-staging/rhel-{dist.replace('el', '')}-x86_64/",
```
ekohl, line 105: Just a thought: can we define `x86_64` as a parameter to the function? I'd love to start supporting arm and possibly other arches for the client repos.

ekohl (commented):
Just a quick note, not a complete review.

build_stage_repository
ehelms: Sure, arch is a parameter now.

build_stage_repository
```diff
@@ -120,16 +124,17 @@ def main():
     try:
         collection = sys.argv[1]
         version = sys.argv[2]
-        operating_system = sys.argv[3]
+        dist = sys.argv[3]
+        arch = sys.argv[4]
```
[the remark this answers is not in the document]
ehelms: I agree that this is a bit trickier with multi-arch, and I'd like to thus tackle that when we take on multi-arch support. At the moment, solving this problem we don't have is blocking my ability to solve the problem we do have and I'd like to keep the momentum going to get Copr support running on nightly. I'll have to solve multi-arch across all of our properties (e.g. Jenkins pipelines and the push to prod script most likely).

I filed https://github.com/theforeman/theforeman-rel-eng/issues/265 to track this for when the time comes.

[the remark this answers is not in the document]
ekohl: For now let's remove the 4th arg and hardcode it `arch = 'x86_64'` for now. At least the rest is then mostly ready. We can deal with `main()` in #265 as you suggest.

ekohl (approved):
Untested, but :+1: from reading the changes.

PR #269, opened 2023-09-06
Add Katello 4.10

chris1984:
@ekohl how does this look?

### DIFF
--- a/releases/katello/4.10/settings
+++ b/releases/katello/4.10/settings
@@ -1,0 +1,4 @@
+FULLVERSION='4.10.0-rc1'
+FOREMAN_VERSION='3.8'
+OSES="el8"
+KOJI_TAGS="katello-@VERSION@-@OS@ katello-candlepin-@VERSION@-el8"


releases/katello/4.10/settings
```diff
@@ -0,0 +1,4 @@
+FULLVERSION='4.10.0'
```
ekohl, line 1: This implies it's GA, but the full version includes the RC part. See dbc04052116c8e58d1323b9894444a0802f5db3f

chris1984: @ekohl fixed, sorry about that I should have looked at a rc1 commit before just blindly copy/pasta

ekohl: For Foreman's branching procedure we have explicit instructions: https://github.com/theforeman/theforeman-rel-eng/blob/5f6403570c60b7144168439562d9b5a3d722de09/procedures/foreman/branch.md.erb#L62

Katello should have a similar step, which would prevent any future confusion. Could you submit a PR?

ekohl (approved):

releases/katello/4.10/settings
chris1984: Yeah will make a follow up today

PR #271, opened 2023-09-06
Fix redmine_version_close for Katello

wbclark:
Katello versions on redmine have names like 'Katello 4.9.2' which prevents an exact match against the version number alone from working.

### DIFF
--- a/procedures/katello/release.md.erb
+++ b/procedures/katello/release.md.erb
@@ -20,7 +20,7 @@
   - [ ] For any cherry-picks that are not needed (including Redmine trackers) you can add them to the `:ignores:` section of `tool_belt` in `configs/katello/<%= short_version %>.yaml`
 <% end -%>
 <% unless is_rc -%>
-- [ ] Change Redmine version <%= full_version %> state to Closed using <%= rel_eng_script('close_redmine_version') %>
+- [ ] Change Redmine version <%= full_version %> state to Closed using <%= rel_eng_script('close_redmine_version') %>: `PROJECT=katello VERSION=<%= short_version %> ./close_redmine_version` and enter your Redmine API Key when prompted.
 <% end -%>
 - [ ] Check for outdated deprecation warnings in the current and next release with `./tools check-deprecation-warnings configs/katello/<%= short_version %>.yaml`. Follow the instructions in the output of the command. Don't forget to create any Redmine issues needed!
 - [ ] Update the `<%= foreman_version %>` branch in [foreman-documentation](https://github.com/theforeman/foreman-documentation)


redmine
```diff
@@ -40,7 +40,7 @@ def set_status_version(args: argparse.Namespace) -> None:
     redmine = get_redmine()
     versions = redmine.version.filter(project_id=args.project_id)
     try:
-        version = [version for version in versions if version.name == args.name][0]
+        version = [version for version in versions if args.name in version.name][0]
```
ekohl, line 43: I'd like to avoid `in` checks with version numbers. You can easily match `3.8.10` when you intended `3.8.1`. How about splitting on space and assuming the last arg is the version?

Is this too hard to read?
```suggestion
        version = [version for version in versions if version.name.split()[-1] == args.name][0]
```

wbclark: Good catch.

I think the nested `[]`s for indexing are a little disorienting for readability in a one-liner; Ruby has spoiled me, where is my good friend `Array#last`?

As an alternative, I can introduce a new method to check something like `if version.name in [args.name, args.project_id + ' ' + args.name]` or some `re` based check.

wbclark:
@ekohl updated, let me know if you like this approach better

procedures/katello/release.md.erb
```diff
@@ -20,7 +20,7 @@
   - [ ] For any cherry-picks that are not needed (including Redmine trackers) you can add them to the `:ignores:` section of `tool_belt` in `configs/katello/<%= short_version %>.yaml`
 <% end -%>
 <% unless is_rc -%>
-- [ ] Change Redmine version <%= full_version %> state to Closed using <%= rel_eng_script('close_redmine_version') %>
+- [ ] Change Redmine version <%= full_version %> state to Closed using <%= rel_eng_script('close_redmine_version') %>: `REDMINE_API_KEY=<secret> PROJECT=katello VERSION=<%= short_version %> ./close_redmine_version`
```
ekohl, line 23: It prompts for `REDMINE_API_KEY` and I'm not sure we should encourage this. The key does end up in your bash history. We can also encourage storing it in gopass, like we do with other secrets.

As for the env vars: that's really what we need for all `rel_eng_script`. I see there's already precedent for it in this release procedure. I tend to use `export PROJECT=... VERSION=...` at the start.

Any thoughts?

redmine
```diff
@@ -36,11 +36,15 @@ def create_version(args: argparse.Namespace) -> None:
         raise SystemExit(str(ex))
 
 
+def match_formatted_version(version_name, args_name, project_id):
+    return version_name == args_name or version_name == (project_id[0].upper() + project_id[1:] + ' ' + args_name)
```
ekohl, line 40: You can at least simplify this to:
```suggestion
    return version_name in (args_name, version_name == (project_id[0].upper() + project_id[1:] + ' ' + args_name))
```

But then you also observe that for every version that's found you construct the string. It would be better to store the potential names in some variable and match against that. See the next comment below:

redmine
```diff
@@ -36,11 +36,15 @@ def create_version(args: argparse.Namespace) -> None:
...
+
+
 def set_status_version(args: argparse.Namespace) -> None:
     redmine = get_redmine()
     versions = redmine.version.filter(project_id=args.project_id)
```
ekohl, line 45: I'd store the matches here. There's a better name for the variable.
```suggestion
    versions = redmine.version.filter(project_id=args.project_id)
    matches = (args.name, f'{args.project_id[0].upper()}{args.project_id[1:]} {args.name})')
```

redmine
```diff
@@ -36,11 +36,15 @@ def create_version(args: argparse.Namespace) -> None:
...
     try:
-        version = [version for version in versions if version.name == args.name][0]
+        version = [version for version in versions if match_formatted_version(version.name, args.name, args.project_id)][0]
```
ekohl, line 47: When you have that variable, it becomes rather simple:
```suggestion
        version = [version for version in versions if version.name in matches][0]
```

ekohl (changes requested):
And I noticed too late that there was another file.

redmine
ekohl: And I realized there's also `capitalize()`, so
```suggestion
    versions = redmine.version.filter(project_id=args.project_id)
    matches = (args.name, f'{args.project_id.capitalize()} {args.name})')
```

ekohl:
I've applied the suggested changes in https://github.com/theforeman/theforeman-rel-eng/pull/296.

ekohl:
> regarding passing sensitive secrets as in-line environment variables, we already do this with `GITHUB_ACCESS_TOKEN=<secret>` when updating changelogs (further below in the same file) so it's consistent with current practice. If we need to make a change there, isn't it larger than the scope of this PR?

I think that's a mistake, rather than something we should continue.

> Personally I'm using `HISTCONTROL=ignorespace` with my shell, so I can prepend a space to any command to prevent it being saved. That said, there are other approaches possible too (storing these secrets in a file somewhere and sourcing it for example) but I'm not sure that we should be prescribing a method here vs. relying on the developer to choose the best approach for their environment.

The script already interactively asks you for it, including the place where you can obtain it:

https://github.com/theforeman/theforeman-rel-eng/blob/940a720b1f530153769d71575eecb54fc3b5a127/close_redmine_version#L5-L9

I think it's better UX if we don't pass it.

wbclark:
Sure, if the script will already prompt for it then I'm fine using that approach. I've pushed the update.

wbclark:
Hey @ekohl , I forgot to `@` you when I pushed the last update... Doing that now to make sure you saw it. Thanks!

ekohl (approved):

PR #272, opened 2023-09-07
Add more precise schedule instructions

### LINKED ISSUE #260: Improve schedule instructions in the Foreman branching procedure
We should provide the exact command to generate the initial schedule at the end of the branching procedure. We already have the date of branching there, so we can easily provide it. The dev start date should be BRANCHING + 1 day.

From there the instructions become a bit more fuzzy: you should look at the dates and see if there are conflicts. Conflicts can be February conferences (FOSDEM and cfgmgmtcamp) (if people involved will attend there). Red Hat timing may also be relevant. While Red Hat shouldn't drive the schedule, the current release team is entirely staffed by Red Hatters. The Red Hat company wide shutdown at the end of the year (between Christmas and New Year) does impact availability of people. Red Hat may also care about Red Hat Summit. It may be good to shift it a week earlier or later. This will never be a hard science.

ekohl:
This provides the exact command to run, removing the need to guess what the day is.

Fixes #260


chris1984 (approved):
ACK, looks good

PR #279, opened 2023-09-25
Update README release procedure to include git pull

chris1984:


chris1984:
@ekohl here is the PR from the action item last week

ekohl (approved):

PR #280, opened 2023-09-29
Switch to building stage repository by combining production and diff …

ehelms:
…of Copr

When the version is nightly, this will not pull from production but instead treat what is in Copr as the source of truth. At the end, the output for a versioned repository is a list of unsigned packages.

There is a lot of change here, and I can likely do some re-factoring now or after merge. I wanted to get a working version of the concept available.

Here is how this generates the stage repository.

For nightly:

 * Copy all RPMs from Copr for the given repository to local
 * Filter the downloaded RPMs through comps file, removing anything not in the comps file
 * Runs createrepo
 
 This is done for RPM and SRPMs. For this workflow, Jenkins will run this script and push the RPMs via rsync to stagingyum.theforeman.org.

For releases:

 * Copy all RPMs from production (yum.theforeman.org) to a local repository
 * Run repodiff to identify new packages in Copr for the release
 * Download the new packages from Copr
 * Filter the downloaded RPMs through comps file, removing anything not in the comps file
 * Runs createrepo
 * Returns as stdout the list of unsigned RPMs and SRPMs
 
For this workflow, the release engineer will run this script, perform a signing of unsigned packages and push the RPMs via rsync to stagingyum.theforeman.org.

The idea is then that either one of these options will happen (via a follow up PR):

 1) New script will take as input the list of unsigned RPMs and sign them
 2) New script will calculate the unsigned RPMs with the location of the repository as input and sign them

### DIFF
--- a/build_stage_repository
+++ b/build_stage_repository
@@ -11,33 +11,19 @@ import hashlib
 import sys
 import tempfile
 import dnf.comps
+import argparse
 
 
-def move_rpms_from_copr_to_stage(collection, version, src_folder, dest_folder, dist, arch, source=False):
-    if source:
-        print(f"Moving {collection} Source RPMs from Copr directory to stage repository")
-    else:
-        print(f"Moving {collection} RPMs from Copr directory to stage repository")
-
-    if not os.path.exists(dest_folder):
-        os.mkdir(dest_folder)
-
-    repo_folder = f"{src_folder}/{dist}-{collection}-{version}-{arch}"
-    packages_to_include = packages_from_comps(comps(collection, version, dist))
-
-    if source:
-        files = glob.glob(repo_folder + f"/**/*.src.rpm")
-    else:
-        files = glob.glob(repo_folder + f"/**/*.rpm")
+def filter_packages(target_dir, packages):
+    files = glob.glob(f"{target_dir}/*.rpm")
 
     for file in files:
         file_name = os.path.basename(file)
         rpm_name = get_rpm_name(file)
 
-        if rpm_name in packages_to_include:
-            shutil.move(file, os.path.join(dest_folder, file_name))
-
-    shutil.rmtree(src_folder)
+        if rpm_name not in packages:
+            print(f"Removed {rpm_name}. Package not in comps.")
+            os.remove(os.path.join(target_dir, file_name))
 
 
 def get_rpm_name(rpm):
@@ -61,18 +47,24 @@ def modulemd_yaml(collection, version):
 
 def comps(collection, version, dist):
     branch = 'develop' if version == 'nightly' else version
+    repo = 'foreman-packaging'
+    name = collection
 
-    if collection == 'foreman-katello':
-        name = 'katello'
-    elif collection == 'foreman-candlepin':
-        name = 'katello-candlepin'
-    else:
-        name = collection
+    if collection == 'client':
+        name = 'foreman-client'
+    elif collection == 'plugins':
+        name = 'foreman-plugins'
+    elif collection == 'candlepin':
+        repo = 'candlepin-packaging'
+    elif collection == 'pulpcore':
+        repo = 'pulpcore-packaging'
 
-    if collection == 'foreman-client' and dist == 'el7':
+    if collection == 'client' and dist == 'el7':
         dist = 'rhel7'
 
-    path, headers = urlretrieve(f"https://raw.githubusercontent.com/theforeman/foreman-packaging/rpm/{branch}/comps/comps-{name}-{dist}.xml")
+    url = f"https://raw.githubusercontent.com/theforeman/{repo}/rpm/{branch}/comps/comps-{name}-{dist}.xml"
+    print(f"Fetching comps: {url}")
+    path, headers = urlretrieve(url)
     return path
 
 
@@ -121,28 +113,32 @@ def create_modulemd(collection, version, stage_dir):
 
 
 def create_repository(repo_dir):
-    check_output(['createrepo', repo_dir])
+    check_output(['createrepo_c', repo_dir], stderr=STDOUT)
 
 
-def sync_copr_repository(collection, version, target_dir, dist, arch, source=False):
-    if source:
-        print(f"Syncing {collection} {version} Source RPM repository from Copr")
+def sync_prod_repository(collection, version, target_dir, dist, arch):
+    repository_url = prod_repository(collection, version, dist, arch)
+
+    if arch == 'source':
+        print(f"Syncing {collection} {version} Source RPM repository from {repository_url}")
     else:
-        print(f"Syncing {collection} {version} RPM repository from Copr")
+        print(f"Syncing {collection} {version} RPM repository from {repository_url}")
 
     cmd = [
         'dnf',
         'reposync',
-        '--newest-only',
+        '--refresh',
+        '--download-metadata',
+        '--norepopath',
         '--repo',
         f"{dist}-{collection}-{version}-{arch}",
         '--repofrompath',
-        f"{dist}-{collection}-{version}-{arch},https://download.copr.fedorainfracloud.org/results/@theforeman/{collection}-{version}-staging/rhel-{dist.replace('el', '')}-{arch}/",
+        f"{dist}-{collection}-{version}-{arch},{repository_url}",
         '--download-path',
         target_dir
     ]
 
-    if source:
+    if arch == 'source':
         cmd.extend([
             '--source',
         ])
@@ -172,28 +168,160 @@ def packages_from_comps(comps):
     return packages
 
 
+def copr_repository(collection, version, dist, arch):
+    return f"https://download.copr.fedorainfracloud.org/results/@theforeman/{collection}-{version}-staging/rhel-{dist.replace('el', '')}-{arch}"
+
+
+def prod_repository(collection, version, dist, arch):
+    if collection == "foreman":
+        subdir = "releases"
+    else:
+        subdir = collection.replace('foreman-', '')
+
+    return f"https://yum.theforeman.org/{subdir}/{version}/{dist}/{arch}"
+
+
+def copr_repository_urls(repository):
+    cmd = [
+        'dnf',
+        'reposync',
+        '--urls',
+        '--refresh',
+        '--repofrompath',
+        f"copr,{repository}",
+        '--repo',
+        'copr'
+    ]
+
+    urls = check_output(cmd, universal_newlines=True, stderr=STDOUT)
+    return urls.split("\n")
+
+
+def compare_repositories(new_repository, old_repository, arch, source=False):
+    cmd = [
+        'dnf',
+        'repodiff',
+        '--simple',
+        '--refresh',
+        '--compare-arch',
+        '--repofrompath',
+        f"new,{new_repository}",
+        '--repofrompath',
+        f"old,{old_repository}",
+        '--repo-old',
+        'old',
+        '--repo-new',
+        'new'
+    ]
+
+    if not source:
+        cmd.extend(['--arch', f'noarch,{arch}'])
+
+    print(' '.join(cmd))
+    return check_output(cmd, universal_newlines=True)
+
+
+def parse_repodiff(diff):
+    split_diff = diff.split("\n")
+    packages = []
+
+    for line in split_diff:
+        if line.startswith('Added package  :'):
+            packages.append(line.replace('Added package  : ', ''))
+        elif ' -> ' in line:
+            packages.append(line.split(' -> ')[1])
+
+    return packages
+
+
+def download_copr_packages(packages, urls, repository, downloads_dir, included_packages):
+    for package in packages:
+        name, version, release = package.rsplit('-', 2)
+
+        if name in included_packages:
+            if ':' in version:
+                version_without_epoch = version.rsplit(':')[1]
+            else:
+                version_without_epoch = version
+
+            name = f"{name}-{version_without_epoch}-{release}.rpm"
+
+            for url in urls:
+                if name == os.path.basename(url):
+                    download_url = url
+
+            if not os.path.exists(f"{downloads_dir}/{name}"):
+                print(f"Downloading {repository}/{name} to {downloads_dir}")
+                urlretrieve(download_url, f"{downloads_dir}/{name}")
+            else:
+                print(f"Skipping {repository}/{name}. Already downloaded.")
+        else:
+            print(f"Skipping {package}. Package not present in comps.")
+
+
+def handle_args():
+    parser = argparse.ArgumentParser(description='Generate a stage repository')
+    parser.add_argument(
+        'collection',
+        help='Repository to generate for (e.g. foreman)'
+    )
+    parser.add_argument(
+        'version',
+        help='Version to generate the repository for (e.g. nightly)'
+    )
+    parser.add_argument(
+        'dist',
+        help='Dist to generate repository for (e.g. el8)'
+    )
+
+    return parser.parse_args()
+
+
+def initialize_repository(collection, version, dist, arch, rpm_dir, srpm_dir):
+    if version != 'nightly':
+        sync_prod_repository(collection, version, rpm_dir, dist, arch)
+        sync_prod_repository(collection, version, srpm_dir, dist, 'source')
+    else:
+        create_repository(rpm_dir)
+        create_repository(srpm_dir)
+
+
+def add_packages_from_copr(collection, version, dist, arch, rpm_dir, srpm_dir):
+    packages_to_include = packages_from_comps(comps(collection, version, dist))
+    copr_repo = copr_repository(collection, version, dist, arch)
+    copr_package_urls = copr_repository_urls(copr_repo)
+
+    rpm_diff = compare_repositories(copr_repo, rpm_dir, arch)
+    packages = parse_repodiff(rpm_diff)
+    download_copr_packages(packages, copr_package_urls, copr_repo, rpm_dir, packages_to_include)
+    filter_packages(rpm_dir, packages_to_include)
+
+    srpm_diff = compare_repositories(copr_repo, srpm_dir, arch, True)
+    srpm_packages = parse_repodiff(srpm_diff)
+    download_copr_packages(srpm_packages, copr_package_urls, copr_repo, srpm_dir, packages_to_include)
+    filter_packages(srpm_dir, packages_to_include)
+
+
+def update_repositories(collection, version, dist, rpm_dir, srpm_dir):
+    create_repository(rpm_dir)
+    create_repository(srpm_dir)
+
+    if collection in ['foreman', 'katello'] and dist == 'el8':
+        create_modulemd(collection, version, rpm_dir)
+
+
 def main():
-    try:
-        collection = sys.argv[1]
-        version = sys.argv[2]
-        dist = sys.argv[3]
-        arch = 'x86_64'
-    except IndexError:
-        raise SystemExit(f"Usage: {sys.argv[0]} collection version os")
+    args = handle_args()
+
+    collection = args.collection
+    version = args.version
+    dist = args.dist
+    arch = 'x86_64'
 
     base_dir = 'tmp'
-    rpm_sync_dir = f"{base_dir}/rpms"
-    srpm_sync_dir = f"{base_dir}/srpms"
-
-    stage_dir = f"{base_dir}/{collection}/{version}/{dist}/"
+    stage_dir = f"{base_dir}/{collection}/{version}/{dist}"
     rpm_dir = f"{stage_dir}/{arch}"
     srpm_dir = f"{stage_dir}/source"
-
-    if not os.path.exists(rpm_sync_dir):
-        os.makedirs(rpm_sync_dir)
-
-    if not os.path.exists(srpm_sync_dir):
-        os.makedirs(srpm_sync_dir)
 
     if not os.path.exists(rpm_dir):
         os.makedirs(rpm_dir)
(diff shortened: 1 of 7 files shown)


ehelms:
I decided to split this into:

 1) Build stage repository with script and new method described in original comment
 2) New script that for a given directory will list unsigned packages
 3) New script that calls list unsigned packages and then feeds them into the sign_rpms script

sign_stage_rpms
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+
+. settings
+
+echo "./list_unsigned_rpms $1 $GPGKEY"
+UNSIGNED_RPMS=(`./list_unsigned_rpms $1 $GPGKEY`)
```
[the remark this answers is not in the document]
ehelms: I don't know what this wants me to do.

upload_stage_rpms
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -e
```
ehelms, line 1: This script is designed to work with this change https://github.com/theforeman/foreman-infra/pull/1948 that allows infra users to rsync as `yumrepostage` user.

list_unsigned_rpms
```diff
@@ -0,0 +1,63 @@
...
+        help='Path to repository to sign unsigned RPMs'
+    )
+    parser.add_argument(
+        'gpgkey',
+        help='Short form gpgkey for the version being generated, does not apply to nightly'
+    )
+
+    args = parser.parse_args()
+
+    if args.gpgkey:
```
ehelms, line 40: Drop this line

list_unsigned_rpms
```diff
@@ -0,0 +1,63 @@
...
+        cmd = [
+            'rpm',
+            '-qpi',
+            package
+        ]
+
+        output = check_output(cmd, universal_newlines=True, stderr=STDOUT)
+
+        if gpgkey.lower() not in output:
+            unsigned.append(package)
```
ehelms, line 22: Can use `yield` to make this a generator

sign_stage_rpms
```diff
@@ -0,0 +1,12 @@
...
+. settings
+
+echo "./list_unsigned_rpms $1 $GPGKEY"
+UNSIGNED_RPMS=$(./list_unsigned_rpms "$1" "$GPGKEY")
+
+for rpm_path in "${UNSIGNED_RPMS[@]}"
+do
+  echo "Signing $rpm_path"
+  ./sign_rpms "$rpm_path"
+done
```
evgeni, line 12: (I hate arrays in bash, never grok them).

This calls `./sign_rpms` once per  unsigned rpm, right? I am asking because `./sign_rpms` will interrogate `gopass` about the password for the key on each invocation and also call `rpm` which can add quite some overhead.

I *think* `./sign_rpms` should be able to handle multiple rpms just fine, so we could either `./sign_rpms $(./list_unsigned_rpms)` or (given that list can get a tad long: `echo $UNSIGNED_RPMS | xargs ./sign_rpms`?

evgeni:
Why do you need to "Copy all RPMs from production (yum.theforeman.org) to a local repository"?
For repodiff, you can just point it at the existing live repo instead.

It it so that you can generate "composed" repo (old plus new) locally, including module metadata? (Today, we push the diff only and the remote regenerates the repodata)

ehelms:
> Why do you need to "Copy all RPMs from production (yum.theforeman.org) to a local repository"? For repodiff, you can just point it at the existing live repo instead.
> 
> It it so that you can generate "composed" repo (old plus new) locally, including module metadata? (Today, we push the diff only and the remote regenerates the repodata)

Yes. This was in service of:

 * the ability to generate a full stage repository locally
 * keep web01 dumb
 * generate module metadata outside of web01 


Thinking about it more I could optimize and have support for that with an option to generate the whole thing locally but require web01 to be a tad smarter than it is today.

### Optimized Workflow

For a release:
 1. Calculate repodiff between production and Copr
 2. Download identified packages from Copr
 3. Filter with comps
 4. Sign packages
 5. Copy signed packages to web01
 6. generate module metadata on web01
 7. update repo metadata

For nightly:

 1. Download all packages from Copr
 2. Filter with comps
 3. Copy all packages to stage
 4. generate module metadata on web01
 5. Createrepo

evgeni:
I think I am cool with the current flow (and generally in favor of dumbing down web01). Just needed to have a clearer picture.

Does downloading the *existing* RPMs from prod keep their filesystem timestamps, or would these be updated during rsync back to web01? If possible, I'd try to retain them, so people have a clearer picture when things changed.

ehelms:
> I think I am cool with the current flow (and generally in favor of dumbing down web01). Just needed to have a clearer picture.
>
> Does downloading the _existing_ RPMs from prod keep their filesystem timestamps, or would these be updated during rsync back to web01? If possible, I'd try to retain them, so people have a clearer picture when things changed.

That will require some testing.

evgeni:
> > I think I am cool with the current flow (and generally in favor of dumbing down web01). Just needed to have a clearer picture.
> > Does downloading the _existing_ RPMs from prod keep their filesystem timestamps, or would these be updated during rsync back to web01? If possible, I'd try to retain them, so people have a clearer picture when things changed.
> 
> That will require some testing.

Tested and test failed.
The following code works tho:
```python
import os
from email.utils import parsedate_to_datetime
from urllib.request import urlretrieve
filename, headers = urlretrieve(…)
if 'Last-Modified' in headers:
    modification_ts = parsedate_to_datetime(headers['Last-Modified']).timestamp()
    os.utime(filename, (modification_ts, modification_ts))
```

build_stage_repository
```diff
@@ -172,40 +166,143 @@ def packages_from_comps(comps):
     return packages
 
 
+def copr_repository(collection, version, dist, arch):
+    return f"https://download.copr.fedorainfracloud.org/results/@theforeman/{collection}-{version}-staging/rhel-{dist.replace('el', '')}-{arch}"
+
+
+def prod_repository(collection, version, dist, arch):
+    if collection == "foreman" and version == "nightly":
+        return f"https://yum.theforeman.org/{version}/{dist}/{arch}"
```
evgeni, line 175:
```suggestion
    if collection == "foreman":
        return f"https://yum.theforeman.org/releases/{version}/{dist}/{arch}"
```

build_stage_repository
```diff
@@ -172,40 +166,143 @@ def packages_from_comps(comps):
...
+    else:
+        return f"https://yum.theforeman.org/{collection.replace('foreman-', '')}/{version}/{dist}/{arch}"
```
evgeni, line 177: or better?
```suggestion
    if collection == "foreman":
        subdir = "releases"
    else:
        subdir = collection.replace('foreman-', '')
    return f"https://yum.theforeman.org/{subdir}/{version}/{dist}/{arch}"
```

build_stage_repository
```diff
@@ -172,40 +166,143 @@ def packages_from_comps(comps):
...
+    packages = parse_repodiff(rpm_diff)
+    download_copr_packages(packages, copr_package_urls, copr_repo, rpm_dir, packages_to_include)
+    filter_packages(rpm_dir, packages_to_include)
+
+    srpm_diff = compare_repositories(copr_repo, srpm_dir, True)
+    srpm_packages = parse_repodiff(srpm_diff)
+    download_copr_packages(srpm_packages, copr_package_urls, copr_repo, srpm_dir, packages_to_include)
+    filter_packages(srpm_dir, packages_to_include)
 
     create_repository(rpm_dir)
```
evgeni, line 307: this will erase "old" module metadata from `rpm_dir` (if there is any, e.g. when `sync_prod_repository` fetched some).

in https://github.com/theforeman/foreman-infra/blob/376d691f2efca2d0bae8eb594c10cfd51e2117f3/puppet/modules/web/files/deploy-yumrepo.sh#L51-L64 there is an example of how we (today) merge old module metadata, and I think this should be somehow transferred here, as otherwise we'll end up with a repo that has RPMs that are not referenced by any version of a module

ehelms: I am not seeing how we would end up with RPMs not in the module metadata. We download production, calculate the diff, download the diff, create a repo and then add module metadata based on all packages in the repository.

ehelms:
> ```python
> import os
> from email.utils import parsedate_to_datetime
> from urllib.request import urlretrieve
> filename, headers = urlretrieve(…)
> if 'Last-Modified' in headers:
>     modification_ts = parsedate_to_datetime(headers['Last-Modified']).timestamp()
>     os.utime(filename, (modification_ts, modification_ts))
> ```

Do I understand correctly:

Your suggestion is to swap away from reposync to a method that downloads the RPMs individually and using this code chunk to do so in order to maintain the modification date.

evgeni:
Nah, just make me read the code better 🙈
The part that fetches prod repo is using reposync, not urlretrieve, that is used only for "new" stuff coming from copr.
Back to square one evgeni.

build_stage_repository
evgeni: Again me mixing up flows. The result feeds in all rpms, not only the new ones.

(What does it tell about me, or the code, that I misread it in two places?)

ehelms:
> Nah, just make me read the code better 🙈

Does the new structure help at all?

build_stage_repository
```diff
@@ -121,28 +113,30 @@ def create_modulemd(collection, version, stage_dir):
...
 
-def sync_copr_repository(collection, version, target_dir, dist, arch, source=False):
-    if source:
-        print(f"Syncing {collection} {version} Source RPM repository from Copr")
+def sync_prod_repository(collection, version, target_dir, dist, arch):
+    if arch == 'source':
+        print(f"Syncing {collection} {version} Source RPM repository from yum.theforeman.org")
     else:
-        print(f"Syncing {collection} {version} RPM repository from Copr")
+        print(f"Syncing {collection} {version} RPM repository from yum.theforeman.org")
```
ekohl, line 123: Would it make sense to determine the URL (which `prod_repository()` does and then print it here?

sign_stage_rpms
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+
+. settings
+
+echo "./list_unsigned_rpms $1 $GPGKEY"
```
ekohl, line 5: As a debug advice: use `bash -xe ./sign_stage_rpms` to make it print exactly what it's doing.
```suggestion
```

And @evgeni will then suggest to change the header to
```bash
#!/bin/bash

set -e
```

Then you can simply use `bash -x ./sign_stage_rpms`.

sign_stage_rpms
ekohl: I'd just use `./list_unsigned_rpms | xargs ./sign_rpms` and avoid `$UNSIGNED_RPMS` altogether. If you want it to print some debug info, make `sign_rpms` echo that it's signing something.

list_unsigned_rpms
```diff
@@ -0,0 +1,57 @@
...
+    args = handle_args()
+
+    repository_path = args.repository_path
+    gpgkey = args.gpgkey
+
+
+    for rpm in find_unsigned_packages(f"{repository_path}/x86_64", gpgkey):
+        print(rpm)
+    for srpm in find_unsigned_packages(f"{repository_path}/source", gpgkey):
+        print(srpm)
```
ekohl, line 53:
```suggestion
    for arch in ('x86_64', 'source'):
        for rpm in find_unsigned_packages(f"{repository_path}/{arch}", gpgkey):
            print(rpm)
```

list_unsigned_rpms
```diff
@@ -0,0 +1,57 @@
...
+import glob
+
+
+def find_unsigned_packages(target_dir, gpgkey):
+    packages = glob.glob(f"{target_dir}/*.rpm")
+
+    for package in packages:
+        cmd = [
+            'rpm',
+            '-qpi',
```
ekohl, line 14: You can only request the signature via the query format:
```suggestion
            '--query',
            '--queryformat', '%{SIGPGP:pgpsig}',
```

list_unsigned_rpms
```diff
@@ -0,0 +1,57 @@
...
+            package
+        ]
+
+        output = check_output(cmd, universal_newlines=True, stderr=STDOUT)
+
+        if gpgkey.lower() not in output:
```
ekohl, line 20: Perhaps we should lower case it in `handle_args`?

build_stage_repository
```diff
@@ -172,47 +166,160 @@ def packages_from_comps(comps):
...
+    for package in packages:
+        name = f"{package}.rpm".replace("-1:", "-")
+
+        for url in urls:
+            if name in url:
+                download_url = url
+
+        if not os.path.exists(f"{downloads_dir}/{name}"):
+            print(f"Downloading {repository}/{name} to {downloads_dir}")
+            urlretrieve(download_url, f"{downloads_dir}/{name}")
```
ekohl, line 244: Does `urlretrieve` keep a session open? If it needs to open a new connection for every file it may be a lot slower.

sign_stage_rpms
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
```
ehelms, line 1: Add an updaterepo step since the RPM checksums will change after signing

list_unsigned_rpms
ehelms: My thinking was to make absolutely sure it's lowered when it's needed to avoid any future changes that might would break this identification of what is signed or not signed.

ehelms:
> Please also modify this as needed:

This is handled over here -- https://github.com/theforeman/theforeman-rel-eng/pull/285. I do not want to update the procedures until all the infrastructure is in place.

ehelms:
@evgeni I was able to add a partial optimization with filtering on download. The reason we cannot do a complete is repodiff returns the full NEVRA, e.g.

```
Added rubygem-pg-debuginfo-1.5.3-1.el8
```

And comps contains just the name `rubygem-pg` so we can at best do a token `in` match of the name.

evgeni:
You can split NEVRAs like this:
```python
  package, _version, _release = nevra.rsplit('-', 2)
```

build_stage_repository
ehelms: I do not know, and some quick searches did not yield any particular result. I've not any particular slowness.

build_stage_repository
```diff
@@ -172,47 +168,167 @@ def packages_from_comps(comps):
...
+
+def download_copr_packages(packages, urls, repository, downloads_dir, included_packages):
+    for package in packages:
+        include = any(included in package for included in included_packages)
+
+        if include:
+            name = f"{package}.rpm".replace("-1:", "-")
+
+            for url in urls:
+                if name in url:
```
ekohl, line 246: I'm wondering about strict matching. Should we at least make it
```suggestion
                if url.endswith(name):
```

I also considered exact matching:
```suggestion
                if name == os.path.basename(url):
```

list_unsigned_rpms
```diff
@@ -0,0 +1,56 @@
...
+        help='Path to repository to sign unsigned RPMs'
+    )
+    parser.add_argument(
+        'gpgkey',
+        help='Short form gpgkey for the version being generated'
+    )
+
+    args = parser.parse_args()
+
+    if len(args.gpgkey) != 8:
```
ekohl, line 39: On Fedora 38 I see:
```console
$ rpm --query --queryformat '%{SIGPGP:pgpsig}' fedora-logos
RSA/SHA256, Mon 20 Feb 2023 19:01:58 CET, Key ID 809a8d7ceb10b464
```

Note how the Key ID is 16 characters long.

generate_stage_repository
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -ex
```
ekohl, line 1: Debug left over?
```suggestion
#!/bin/bash -e
```

If it's intentional, I'd instead suggest to enable it after loading settings

```bash
set -x
for os in $OSES; do
  ./build_stage_repository "$PROJECT" "$VERSION" "$os"
done
set +x
```

build_stage_repository
```diff
@@ -172,47 +168,167 @@ def packages_from_comps(comps):
...
+        '--repofrompath',
+        f"old,{old_repository}",
+        '--repo-old',
+        'old',
+        '--repo-new',
+        'new'
+    ]
+
+    if not source:
+        cmd.extend(['--arch', 'noarch,x86_64'])
```
ekohl, line 218: In other places we have the arches as parameters. Just wondering, would it be
```suggestion
        cmd.extend(['--arch', f'noarch,{arch}'])
```

And then iterate that for every arch.

build_stage_repository
ekohl: I was trying to understand if this maintains a persistent connection between downloads so I went to the docs: https://docs.python.org/3/library/urllib.request.html#urllib.request.urlretrieve

First thing to notice is that it's listed as a legacy method. It also links to `urlopen()` which states:

> urllib.request module uses HTTP/1.1 and includes Connection:close header in its HTTP requests.

So it closes the connection after fetching the file, which means you set up a lot of connections, which has a certain overhead. When I ping the endpoint I have 6ms latency, so it probably isn't that big of a deal. Possible future optimization.

list_unsigned_rpms
ehelms: Aye, with your method for the query the full length key is now available. Will update.

ehelms:
I noticed that rpmsign does seem to help prevent us from re-signing an already signed package:

```
warning: tmp/foreman/3.8/el8/source/rubygem-unf_ext-0.0.8.2-1.el8.src.rpm already contains identical signature, skipping
```

ehelms:
Updated with some included README workflow layout. This can be merged prior to (https://github.com/theforeman/foreman-infra/pull/1948) for nightly, but https://github.com/theforeman/foreman-infra/pull/1948 must be merged prior to this if I am to test this for releases and rsyncing to stagingyum.

evgeni:
One thing that I realized while reviewing https://github.com/theforeman/foreman-infra/pull/1948:
Today, nothing cleans up `~yumrepostage/rsync_cache` and the script that copies over the cache to the final destination doesn't use `--delete`. This is fine™, as we're aiming at a "merged" workflow for the release repos anyway, but it also means that the nightly repo might grow and not get cleanup (at least that's how I read the code). We could add a cleanup step once things get copied to the "final destination", but then the `rsync` from the release engineer workstation will be longer (as it will have to transfer the already signed files again).

Shouldn't block this PR, but something to keep an eye on.

list_unsigned_rpms
```diff
@@ -0,0 +1,56 @@
...
+
+def handle_args():
+    parser = argparse.ArgumentParser(description='Sign unsigned RPMs in local stage repository')
+    parser.add_argument(
+        'repository_path',
+        help='Path to repository to sign unsigned RPMs'
+    )
+    parser.add_argument(
+        'gpgkey',
+        help='Short form gpgkey for the version being generated'
```
evgeni, line 34: I wonder, why not pass the "full" key and let this script adjust the length as it sees fit?

build_stage_repository
```diff
@@ -172,47 +168,167 @@ def packages_from_comps(comps):
...
+    return packages
+
+
+def download_copr_packages(packages, urls, repository, downloads_dir, included_packages):
+    for package in packages:
+        name, _version, _release = package.rsplit('-', 2)
+
+        if name in included_packages:
+            # Remove the Epoch as it does not appear in the URL
+            name = f"{package}.rpm".replace("-1:", "-")
```
evgeni, line 243: instead of replacing `-1:`, you could take the `_version` from aboves `rsplit`, split it at `:` and only use the second part?

list_unsigned_rpms
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    args = parser.parse_args()
+
+    if len(args.gpgkey) != 16:
+        raise SystemExit("GPG key must be the last 16 characters")
```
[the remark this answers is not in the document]
ehelms: `must be` == `exactly`

list_unsigned_rpms
ehelms: I didn't want this script to "interpret" anything and get it wrong. You tell it what the key is you want it to compare against.

generate_stage_repository
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+. settings
+
+set -x
+
+for os in $OSES; do
+  ./build_stage_repository "$PROJECT" "$VERSION" "$os"
```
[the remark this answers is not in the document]
ehelms: I used to expect `foreman-katello`, I have since changed that to what you said in the first line.

build_stage_repository
ehelms: > instead of replacing `-1:`, you could take the `_version` from aboves `rsplit`, split it at `:` and only use the second part?

(Is there a clever emoji?)
(ah, yes: :de: )

generate_stage_repository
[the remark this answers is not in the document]
evgeni: not here:
https://github.com/theforeman/theforeman-rel-eng/blob/216976b94c0dbcd7ce4044849ad72d94b4aeac10/build_stage_repository#L305-L306

list_unsigned_rpms
evgeni: fair.

generate_stage_repository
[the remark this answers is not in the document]
ehelms: There is always one stray.

build_stage_repository
ehelms: See if this update captures it.

evgeni: yeah

evgeni (approved):

build_stage_repository
```diff
@@ -172,47 +168,171 @@ def packages_from_comps(comps):
...
+
+def download_copr_packages(packages, urls, repository, downloads_dir, included_packages):
+    for package in packages:
+        name, version, release = package.rsplit('-', 2)
+
+        if name in included_packages:
+            if ':' in version:
+                version_without_epoch = version.rsplit(':')[1]
+            else:
+                version_without_epoch = version
```
ekohl, line 245: Is this too smart? `version_without_epoch = version.rsplit(':', 1)[-1]`
https://github.com/theforeman/theforeman-rel-eng/pull/286

PR #283, opened 2023-10-11
Wait for Foreman GA before announcing Katello

ianballou:


chris1984 (approved):

ekohl (commented):
I think we should do this before we run the release pipeline so we also test it against GA instead of the RC. Perhaps even before we start packaging so the RPMs are built using the GA. This can be a difference if there are changes around webpack.

ianballou:
@ekohl I changed the PR to have two waiting points on the Foreman GA.

ekohl (approved):
Thanks!

PR #288, opened 2023-10-31
support multi-module modules.yaml

evgeni:


build_stage_repository
```diff
@@ -99,15 +99,21 @@ def create_modulemd(collection, version, stage_dir):
...
+        modules = list(yaml.safe_load_all(file))
+
+    module_version = generate_modulemd_version(version)
+    module_context = generate_modulemd_context(collection, version)
+
+    for module in modules:
+        module['data']['version'] = module_version
+        module['data']['context'] = module_context
+        if module['data']['name'] == collection:
+            module['data']['artifacts'] = {'rpms': output.splitlines()}
```
evgeni, line 111: I want to set the version and context for all modules in the file, but artifacts only for the one that is actually our "main" module

ekohl (approved):

PR #291, opened 2023-11-01
Add nightly for each repository release stream

ehelms:
This change will allow controlling for what OSes a given release from our release jobs, for nightly, and align it with a numerical release. Right now, we store in Jenkins the OSes nightly supports and then we iterate and call the `build_stage_repository` script [1]. By having the nightly releases captured here, our CI can instead call `VERSION='nightly' PROJECT=<project> ./generate_stage_repositories` and allow configuration inside this repository to drive what is generated.

In a similar fashion, we can make use of a single rsync script to stagingyum to handle uploads for nightly and releases in the same way without having to keep CI and this repository in sync. See https://github.com/theforeman/jenkins-jobs/pull/369.

[1] https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/release/pipelines/foreman-rpm.groovy#L69-L73


ekohl (changes requested):
We don't have rc1 in nightly. I'm not sure what a valid `FULLVERSION` would be. For example, in our release pipeline we pass `FULLVERSION` as the expected version and that would mean it expects `nightly.0-rc1` in `/usr/share/foreman*/VERSION`. Perhaps it should just not be set at all?

ehelms:
> We don't have rc1 in nightly. I'm not sure what a valid `FULLVERSION` would be. For example, in our release pipeline we pass `FULLVERSION` as the expected version and that would mean it expects `nightly.0-rc1` in `/usr/share/foreman*/VERSION`. Perhaps it should just not be set at all?

For now I shortened it to nightly. Right now this appears to be forced by https://github.com/theforeman/theforeman-rel-eng/blob/master/settings#L25-L29 and I wouldn't want to relax this for releases. Whereas having this ability to use scripts in all versions including nightly is strong for centralizing our scripts. Perhaps we take this version as is and keep iterating if FULLVERSIOn for nightly becomes tricky with other workflows?

ekohl:
In https://github.com/theforeman/theforeman-rel-eng/pull/330 I made an effort to make some variables optional, which makes the files more logical.

ekohl (commented):
I think we can now drop `FULLVERSION` in all of these files. The flows that need it shouldn't run and otherwise properly error out.

ehelms:
> I think we can now drop `FULLVERSION` in all of these files. The flows that need it shouldn't run and otherwise properly error out.

Cleaned up, and seems to be working from my test scenarios on scripts.

releases/client/nightly/settings
```diff
@@ -0,0 +1 @@
+FOREMAN_VERSION='nightly'
```
ekohl, line 1: We set this by default:
https://github.com/theforeman/theforeman-rel-eng/blob/6a9a8a39e3b0a4889102c7add4132a624a1ef5cb/releases/client/settings#L3
```suggestion
```

releases/candlepin/nightly/settings
```diff
@@ -0,0 +1 @@
+FOREMAN_VERSION='nightly'
```
ekohl, line 1: Candlepin doesn't have a Foreman version:
https://github.com/theforeman/theforeman-rel-eng/blob/6a9a8a39e3b0a4889102c7add4132a624a1ef5cb/releases/candlepin/settings#L4

```suggestion
```

releases/plugins/nightly/settings
```diff
@@ -0,0 +1 @@
+FOREMAN_VERSION='nightly'
```
ekohl, line 1: I'd prefer a solution similar to `client` where in `releases/plugins/settings` we set `FOREMAN_VERSION=$VERSION` and be done with it.
```suggestion
```

releases/plugins/settings
```diff
@@ -0,0 +1,2 @@
+PACKAGING_SUBDIR="packages/plugins"
+OSES="el8"
```
ekohl, line 2: As said above: we can set it by default (and sort the file alphabetically):
```suggestion
FOREMAN_VERSION=$VERSION
OSES="el8"
PACKAGING_SUBDIR="packages/plugins"
```

ekohl (changes requested):
I should have caught this in the previous review, but more simplifications are possible.

ekohl (changes requested):
Wouldn't you still need a `release/plugins/nightly/settings` file, since it will error out if it doesn't exist? By now that needs `OSES="el8 el9"`.

releases/pulpcore/nightly/settings
```diff
@@ -0,0 +1 @@
+FULLVERSION='nightly.0-rc1'
```
ekohl, line 1: Pulpcore doesn't have a full version either:
```suggestion
```

ekohl (changes requested):
Last one, I promise ;)

releases/client/settings
```diff
@@ -1,3 +1,4 @@
+FOREMAN_VERSION=$VERSION
```
evgeni, line 1: This line now exists twice in this file

evgeni (changes requested):
Glad *I* didn't promise anything

releases/client/settings
ehelms: Snuck right in there! Also reading comments out of order is both confusing and fun.

ekohl (approved):

PR #292, opened 2023-11-01
Use explicit rsync options to detect unchanged RPMs when uploading to stagingyum

ehelms:
I am proposing we use `checksum` to determine if an RPM should get synced to stagingyum.theforeman.org. In the current state, using the default of mtime and size, we download the packages from Copr and they all end up getting synced to stagingyum (and thus to yum.theforeman.org) whether they changed or not.

@evgeni @ekohl This is a follow on to our bigger original discussion around not overriding RPMs in production that have not changed.

### DIFF
--- a/upload_stage_rpms
+++ b/upload_stage_rpms
@@ -5,4 +5,4 @@
 USER='yumrepostage'
 HOST='web01.osuosl.theforeman.org'
 
-rsync --archive --verbose --partial --one-file-system --delete-after "tmp/$PROJECT/$VERSION" "$USER@$HOST:rsync_cache/$PROJECT"
+rsync --checksum --perms --recursive --links --verbose --partial --one-file-system --delete-after "tmp/$PROJECT/$VERSION" "$USER@$HOST:rsync_cache/$PROJECT"


evgeni:
Ooh, I forgot to tell you…

I did some experiments (outside the scripts here), and it seemed to me that (at least on Fedora 38) it doesn't matter which "is file changed" logic you select and the modification time of a file is always kept uptodate when you use `--times`.

"But I don't use `--times`!" you'll say. Well, but `--archive`, which is defined as
```
-a, --archive               archive mode; equals -rlptgoD (no -H,-A,-X)
```
And `-t` is `--times`.

In my tests, switching `--archive` to the more arcane, verbose notation and dropping `t` from it made the files retain their original mtimes.

Yes I wish there would be `--archive --no-times`, but there isn't.

ehelms:
Excellent. I've updated this to reflect that idea and spell out each option with verbosity. I also found with my local testing the same result.

ehelms:
Were you saying to drop `--archive`, trim the list of what we use from it but also use --checksum as our change detection?

ehelms:
> Yeah, you still need --checksum (or --size-only, but I prefer checksum) to avoid "different timestamp, file changed"

I did a few laps around the round-about but I think I finally exited properly.

ehelms:
Note that this change will be needed in CI as well, however, I am proposing to re-use this script in CI to avoid the two-places problem -- https://github.com/theforeman/jenkins-jobs/pull/369

evgeni:
> Were you saying to drop `--archive`, trim the list of what we use from it but also use --checksum as our change detection?

Yes.

evgeni (approved):

PR #295, opened 2023-11-06
Set Foreman version in Katello release procedure

ekohl:
Previously this had a placeholder FOREMAN_VERSION, dating back from when it was in tool_belt. Now everything is in the same repository, this can be provided with a correct value.


ianballou (approved):
Working for me!

PR #304, opened 2023-11-16
Add Pulpcore 3.39 config

Odilhao:


ekohl (approved):
Confirmed I can import the private key.

PR #321, opened 2023-12-05
Update generate stage script to use the Foreman Version

Odilhao:


build_stage_repository
```diff
@@ -275,7 +281,10 @@ def handle_args():
         'dist',
         help='Dist to generate repository for (e.g. el8)'
     )
-
+    parser.add_argument(
+        'foreman_version',
```
ehelms, line 285: Can we make this a `--` option since it only applies to Katello?

build_stage_repository
```diff
@@ -39,13 +39,17 @@ def get_rpm_name(rpm):
     return check_output(command, universal_newlines=True)
 
 
-def modulemd_yaml(collection, version):
+def modulemd_yaml(collection, version, foreman_version):
```
ehelms, line 42: Can we make this here, and in other places, have a default value of None? To match my question around having a `--foreman-version` option?

ehelms (approved):

PR #330, opened 2023-12-18
Reduce settings needed for every release

ekohl:
This explicitly requires FULLVERSION to be set when needed. This allows them to be dropped from client and pulpcore settings.

It also automatically sets it for `FOREMAN_VERSION`, not in every single settings file.


ekohl:
Untested for now, so a draft.

@ehelms is `FULLVERSION` needed to be set for Candlepin, or can it be dropped like Pulpcore?

ehelms (approved):
Looks sane to me.

ehelms:
@ekohl I'd love for you to revisit this so I can finish off the Copr work and align our use of scripts in CI and locally.

ekohl:
@evgeni mind giving this a sanity check?

evgeni (approved):
I did not check if there are *other* scripts that expect $FULLVERSION and didn't get the `require_fullversion` treatment, but the diff here in the PR looks sane to me!

PR #332, opened 2023-12-19
Add "fixed in" update to Katello CP procedure

ianballou:
Since we don't have automation for keeping the fixed in field updated past when a PR is merged, the release owner needs to manually update the "Fixed in Releases" field to include the version receiving the cherry-picks.


ekohl (changes requested):
As discussed in today's meeting: the automation should take care of this. https://github.com/theforeman/prprocessor/pull/176 was intended to do exactly that. I'm not sure if I made a mistake, but let's first investigate that before adding manual steps.

PR #335, opened 2024-01-05
preserve times when generating and uploading staging repo

evgeni:


build_stage_repository
```diff
@@ -256,7 +258,10 @@ def download_copr_packages(packages, urls, repository, downloads_dir, included_p
 
             if not os.path.exists(f"{downloads_dir}/{name}"):
```
ekohl, line 259: One risk I see: people may still have local files cached that won't be updated, causing it to flip back & forth. Nothing you can do about it here, but one we can solve by communicating it.

cc @theforeman/release-engineering

ekohl (approved):

Issue #336, opened 2024-01-08
Newbie installation issues


Griffin-Sullivan:
I think it'll be easier to just run through this list and check them off or delete them as we fix.

- [ ] Tag_project needed me to assign a key for gpg git. Something like git config --global user.signingkey <my_key>
- [ ] Tag_push didn’t work because tool_belt used http clone had to reclone everything with ssh
- [ ] Bump_packaging needs foreman-packaging cloned into ~/dev and run the following:

```
Pip install obal
Yum install git-annex rpmdev-tools
In ~/.gitconfig
[alias]
  ph = "!git push $(whoami) HEAD --set-upstream --quiet"
  pr = "!hub pull-request"
  phr = "!git ph && git pr"
```
- [ ] Drop Append CLI release notes taken from the [hammer-cli](https://github.com/theforeman/hammer-cli/blob/3.6-stable/doc/release_notes.md) and [hammer-cli-foreman](https://github.com/theforeman/hammer-cli-foreman/blob/3.6-stable/doc/release_notes.md) changelogs, in [theforeman.org](https://github.com/theforeman/theforeman.org/trees/gh-pages/_includes/manuals/3.6/1.2_release_notes.md). From the instructions
- [ ] Add contributors script to theforeman-rel-eng
- [ ] Ctrl + d escapes script for release_announcement
- [ ] Specify that updates to the /web folder are done in master only
- [ ] Can’t push develop because of checks
- [ ] Can’t push directly to foreman-documentation master
- [ ] Update web/content/index.adoc and web/content/js/versions.js to add 3.9 as a release candidate” Shouldn’t exist
- [ ] Add RC tester role
- [ ] Close_redmine_version: Failed to require python-redmine
- [ ] Add “in a PR” to changing documentation to supported / unsupported
- [ ] Versions update should also add old release as unsupported
- [ ] Update full version before generating the release process 
- [ ] jenkins.JenkinsException: Could not parse JSON info for job when running jenkins related processes

ekohl:
> * Tag_project needed me to assign a key for gpg git. Something like git config --global user.signingkey <my_key>

I don't think this is tracked before, but certainly something that belongs in the README

> * Tag_push didn’t work because tool_belt used http clone had to reclone everything with ssh

I've configured git to rewrite push URLs: `git config --global url.git@github.com:.pushInsteadOf https://github.com/`

> * Bump_packaging needs foreman-packaging cloned into ~/dev and run the following:

[...]

PR #337, opened 2024-01-10
Avoid the need to have git phr as a configured alias

ekohl:
I'm not 100% sure this is better or worse, but it needs some solution. Either we document the aliases or finish this.

Came up in https://github.com/theforeman/theforeman-rel-eng/issues/336

### DIFF
--- a/bump_packaging
+++ b/bump_packaging
@@ -51,7 +51,9 @@ fi
 git commit -m "Release $FULLVERSION"
 
 if [[ $PACKAGING_PR == true ]] ; then
-	git phr -m "Release $FULLVERSION" -b "$REMOTE_BRANCH"
+	# TODO: assumes a remote named whoami for the user's fork
+	git push --set-upstream "$(whoami)" HEAD
+	hub pull-request -m "Release $FULLVERSION" -b "$REMOTE_BRANCH"
 else
 	git push "$PACKAGING_GIT_REMOTE" "HEAD:$REMOTE_BRANCH"
 fi


bump_packaging
```diff
@@ -51,7 +51,9 @@ fi
 git commit -m "Release $FULLVERSION"
 
 if [[ $PACKAGING_PR == true ]] ; then
-	git phr -m "Release $FULLVERSION" -b "$REMOTE_BRANCH"
+	# TODO: assumes a remote named whoami for the user's fork
```
wbclark, line 54: If you really want to script this and reduce the amount of assumptions and manual set-up, our team has been using this ansible role for well over a year to deploy all of the source and tooling repositories used for building hotfix RPMs: https://github.com/theforeman/forklift/pull/1580

I've been using it to manage all of the repositories in my katello devel box for a similar length of time as well.

ekohl: I'm not a fan of using VMs for those common tasks. It's always awkward when I miss my own customization.

bump_packaging
```diff
@@ -51,7 +51,9 @@ fi
...
+	git push --set-upstream "$(whoami)" HEAD
+	hub pull-request -m "Release $FULLVERSION" -b "$REMOTE_BRANCH"
```
evgeni, line 56: I've been using [`gh`](https://github.com/cli/cli) and there it's `gh pr create` and it will take care of pushing and everything.

ekohl: Somehow I thought `gh` was written in Javascript and always ignored it, but now I see it's not and it's even packaged in Fedora. Maybe I'll take some time to look at converting.

evgeni: Yeah, I'm using the Fedora package and it's quite OK for the things I needed it for.

ekohl: I noticed you often push to the original repository instead of your own fork, which is something I'd like to avoid. Is that because of `gh` behavior?

evgeni: No, that's my lazyness, you can tell it where to push to. I just often don't even have an own fork (configured)

ogajduse:
I'd say this is no longer applicable as #521 was merged.

PR #338, opened 2024-01-25
Allow branching katello without foreman_version

wbclark:
This was previously added as a requirement because of the Katello release procedure, but it's not used anywhere in the Katello branching procedure currently, and the too-strict conditional is breaking the command that generates the Katello branching procedure.

### DIFF
--- a/procedure
+++ b/procedure
@@ -133,7 +133,7 @@ if procedures.key?(procedure)
   command = procedures[procedure]
   command.order!
 
-  unless context[:project] && context[:target_date] && context[:owner] && (context[:project] != 'katello' || context[:foreman_version])
+  unless context[:project] && context[:target_date] && context[:owner] && (context[:project] != 'katello' || procedure != 'release' || context[:foreman_version])
     STDERR.puts command.help
     exit 1
   end


ekohl:
I'll need to take a look. Katello branching still relies on Foreman branching. Many critical steps (like packaging) are now in the Foreman procedure.

On mobile now, so hard to check but do we call these sychronization steps out? For release we do say not to announce GA before Foreman, but perhaps these should be called out more

wbclark:
> I'll need to take a look. Katello branching still relies on Foreman branching. Many critical steps (like packaging) are now in the Foreman procedure.
> 
> On mobile now, so hard to check but do we call these sychronization steps out? For release we do say not to announce GA before Foreman, but perhaps these should be called out more

I looked for content in the katello branching template that used the foreman_branching variable, or for obvious places where it would benefit from that addition, and couldn't find any.

The key issue is that `context[:foreman_version]` is not set in the command corresponding to `branch` in the `procedures` hash (but is set for the `release` command by comparison), so the `unless` conditional was blocking @qcjames53 and I from rendering the template.

I don't mind setting `context[:foreman_version]` in the `branch` command either to fix this, I just didn't see a need for it today based on the current template contents.

ekohl (approved):
We indeed don't pass it in here:
https://github.com/theforeman/theforeman-rel-eng/blob/fbe0aabfd23136a46bbf33e4186a38de483b9b42/procedure_branch#L25

PR #340, opened 2024-01-26
Use branching date as target_date in branching procedures

wbclark:
All procedures currently accept a single target_date. For branching procedures in particular, if the branching date is used as the target_date, this generates correct dates in the section headings but an incorrect date in the command invocation to generate the release procedure. This commit fixes the inconsistency by assuming the branch date is the target_date in branching, and adding four weeks to that date to get the date for GA.

### DIFF
--- a/procedures/foreman/branch.md.erb
+++ b/procedures/foreman/branch.md.erb
@@ -121,7 +121,7 @@ The next step should only be done after all branching, including packaging, has 
 ## Release Owner
 
 - [ ] Generate and post the release procedure, if not already posted:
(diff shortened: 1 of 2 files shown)


ekohl (changes requested):
This doesn't look right. We publish this procedure on the target date and then release Foreman X.Y.0-rc1 on the same day. We also do the same with Katello.

ekohl:
That is exactly what is happening, so it's all working as intended. Targeted branching date is equal to targeted RC1 date.

wbclark:
Cool, thanks for the explanation.

PR #341, opened 2024-01-29
Add missing backtick to make it a command

ekohl:
Fixes: 04bf57552482 ("Add a rel_eng_script macro in procedures")


wbclark (approved):
Heads up @qcjames53

ekohl:
I already modified the existing post, which is where I noticed it.

PR #347, opened 2024-02-21
Update README to include signing key config

Griffin-Sullivan:


README.md
```diff
@@ -114,6 +114,12 @@ For running jobs in Jenkins, make sure you have [access](https://theforeman.gith
 gopass edit theforeman/jenkins-token --create
 ```
 
+For pushing tags in theforeman repositories, you will need to sign your commits. To do so, add your gpg key id as your signing key for git by running:
```
ekohl, line 117: I'd move this under the general `Setup` section, not `Gopass`. It's also not pushing tags to the repos, but rather creating signed tags. Pushing is the step beyond that. Perhaps something like:
```suggestion
Foreman requires GPG signed git tags. Configure git with your personal gpg key id as your signing key:
```

ekohl (changes requested):

PR #348, opened 2024-02-22
Extend PR template after branching docs

maximiliankolb:
In "foreman-documentation", there are a lot of open upstream PRs and there is typically a lot of cherry-picking to release branches. When we push the new release branch to foreman-documantation, we should immediately add this as a friendly reminder to a) users opening new PRs and b) maintains that cherry-pick PRs to the PR template.

cc @Griffin-Sullivan 

Related to https://github.com/theforeman/foreman-documentation/pull/2821 and https://github.com/theforeman/foreman-documentation/issues/2820

### DIFF
--- a/procedures/foreman/branch.md.erb
+++ b/procedures/foreman/branch.md.erb
@@ -88,6 +88,7 @@
 - [ ] Create <%= release %>-stable branches using <%= rel_eng_script('branch_project') %>
 - [ ] Push <%= release %>-stable branches using <%= rel_eng_script('branch_push') %>
 - [ ] Branch foreman-documentation git repository and update website for new release by following the [README instructions](https://github.com/theforeman/foreman-documentation#branching-new-release)
+- [ ] Create a tiny PR in foreman-documentation to add `<%= release %>` to [`.github/PULL_REQUEST_TEMPLATE.md`](https://github.com/theforeman/foreman-documentation/blob/master/.github/PULL_REQUEST_TEMPLATE.md)
 - [ ] Ask Hammer maintainer to branch [hammer-cli](https://github.com/theforeman/hammer-cli) and [hammer-cli-foreman](https://github.com/theforeman/hammer-cli-foreman)
 
 The next step should only be done after all branching, including packaging, has completed.


procedures/foreman/branch.md.erb
```diff
@@ -88,6 +88,7 @@
 - [ ] Create <%= release %>-stable branches using <%= rel_eng_script('branch_project') %>
 - [ ] Push <%= release %>-stable branches using <%= rel_eng_script('branch_push') %>
 - [ ] Branch foreman-documentation git repository and update website for new release by following the [README instructions](https://github.com/theforeman/foreman-documentation#branching-new-release)
+- [ ] Create a tiny PR in foreman-documentation to add `<%= release %>` to [`.github/PULL_REQUEST_TEMPLATE.md`](https://github.com/theforeman/foreman-documentation/blob/master/.github/PULL_REQUEST_TEMPLATE.md)
```
Griffin-Sullivan, line 91: In the line above it says to do this as part of the foreman-documentation branching process here: https://github.com/theforeman/foreman-documentation#branching-new-release. You can see I made that change in https://github.com/theforeman/foreman-documentation/pull/2803/files. I guess part of the reason you're commenting this is because we don't always do a good job of closing the branching PR before starting the release updates so it gets added into a mega PR.

maximiliankolb: I believe we were too slow to review it. But because there's a lot of cherry-picking in foreman-documentation, I believe it makes sense to add a single line "cherry-pick to _NEW FOREMAN RELEASE_?" to the PR template immediately after branching.

ekohl: It's already part of https://github.com/theforeman/foreman-documentation/commit/fd460a01e91cd64678a945f1f82cda0f9d8dbff7. Why does it need to be here too?

The idea is that we don't duplicate exact branching instructions here, but rather refer to the authoritative source, which in this case is foreman-documentation.

For reference, https://github.com/theforeman/jenkins-jobs/blob/master/branch-foreman is a script where there's more automated.

Another design idea behind this procedure is that the one executing it constantly reflects on the process and improves where needed. So if foreman-documentation is painful, it's good to automate more.

ekohl:
@maximiliankolb given it's already part of the foreman-documentation steps which are referred to, can we close this?

PR #349, opened 2024-02-27
create empty repository when reposync from prod fails

evgeni:
when there is no prod repo, reposync will fail. this seems to be OK on EL7 and Fedora, but the DNF in EL8 exits non-zero in this case and this interrupts the stage repository building.
just catching this error is not sufficient, as then repodiff will fail on the non-existing repository, so we have to create an empty one.

### DIFF
--- a/build_stage_repository
+++ b/build_stage_repository
@@ -159,7 +159,11 @@ def sync_prod_repository(collection, version, target_dir, dist, arch):
             '*.src',
         ])
 
-    check_output(cmd)
+    try:
+        check_output(cmd)
+    except CalledProcessError:
+        print("reposync failed, creating empty repository")
+        create_repository(target_dir)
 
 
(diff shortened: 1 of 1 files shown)


evgeni:
The target directory does never exist, as it's running the job in a clean environment.

evgeni:
Also, to be honest. That's how it worked until today, where we used non-EL8 hosts to run this job.

ekohl:
Is there a scenario where reposync would partially sync something and this would mask it?

evgeni:
most probably yes

evgeni:
The "real" fix would be to set `skip_if_unavailable=True` on these repos, but I have no idea how to do this on the command line.

evgeni:
https://github.com/theforeman/theforeman-rel-eng/pull/350

PR #350, opened 2024-02-27
ignore non-available repositories when syncing prod

evgeni:
when there is no prod repo, reposync will fail otherwise. same applies to the repodiff step later.

This is not a problem on EL7 and Fedora, as those set `skip_if_unavailable=True` by default, but EL8+ sets it to `False`.

### DIFF
--- a/build_stage_repository
+++ b/build_stage_repository
@@ -136,6 +136,8 @@ def sync_prod_repository(collection, version, target_dir, dist, arch):
 
     cmd = [
         'dnf',
+        '--config',
+        f'tmp/dnf.conf',
         'reposync',
         '--refresh',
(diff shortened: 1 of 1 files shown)


build_stage_repository
```diff
@@ -136,6 +136,8 @@ def sync_prod_repository(collection, version, target_dir, dist, arch):
 
     cmd = [
         'dnf',
+        '--config',
+        f'tmp/dnf.conf',
```
evgeni, line 140: this should be `{base_dir}/dnf.conf` but we don't pass `base_dir` to those functions and I was lazy to make it so

ekohl (approved):

PR #352, opened 2024-03-01
Add docs in schedule

ekohl:


ekohl:
I was reading up on this and found https://projects.theforeman.org/projects/foreman/wiki/Release_Management which actually takes the GA date as leading, rather than branching.

chris1984 (approved):
ACK

PR #355, opened 2024-03-05
Foreman 3.10.0 rc2

pcreech:


releases/client/3.10/settings
```diff
@@ -1 +1,2 @@
-OSES="el7 el8 el9"
+FULLVERSION='3.10.0-rc2'
```
evgeni, line 1: Client doesn't need that

evgeni (changes requested):

releases/client/3.10/settings
pcreech: 'twas an artifact of branching, removing

releases/client/3.10/settings
```diff
@@ -1 +1 @@
-OSES="el7 el8 el9"
+OSES='el7 el8 el9'
```
evgeni, line 1: Pedantic, but why the quote change?

pcreech: because locally i had ''' instead of '"', noticed just now as well, fixing, as technically it should be unmodified with the removal of the foreman version

pcreech: Should be fixed now to only change 1 file.

evgeni (approved):

PR #360, opened 2024-03-25
Implement a build_stage_repository wrapper

ekohl:
This wraps the logic that is currently in jenkins-jobs. It uses the fact the repository list is already in the settings and then iterates over it.

Currently untested, writing the jenkins-jobs logic that builds on it now.

### DIFF
--- a/build_stage_repositories
+++ b/build_stage_repositories
@@ -1,0 +1,7 @@
+#!/bin/bash -e
+
+. settings
+
+for OS in $OSES ; do
+	./build_stage_repository "$PROJECT" "$VERSION" "$OS"
+done


ehelms:
I'd like you to meet https://github.com/theforeman/theforeman-rel-eng/blob/master/generate_stage_repository :) My plan all along has been to use scripts from this repo (see https://github.com/theforeman/theforeman-rel-eng/pull/291#issue-1972624558) and https://github.com/theforeman/jenkins-jobs/pull/369 is the first update I want to make before proceeding on to more scripts.

ekohl:
Somehow I completely missed the existence of that script. I was playing around with reviewing your PR and reason about it.

ekohl:
Btw, the naming of [generate_stage_repository](https://github.com/theforeman/theforeman-rel-eng/blob/master/generate_stage_repository) is IMHO misleading since it's generating multiple repositories, so should be plural instead of singular.

ehelms:
> Btw, the naming of [generate_stage_repository](https://github.com/theforeman/theforeman-rel-eng/blob/master/generate_stage_repository) is IMHO misleading since it's generating multiple repositories, so should be plural instead of singular.

Fair-ish, it could generate one or more repositories! Depends on the definition of OSES :)

PR #363, opened 2024-04-29
clean up after ContentTooShortError

evgeni:
sometimes urlretrieve will not get the whole file and raise a ContentTooShortError. it will not delete the partially downloaded file in this case, which in a later run will cause issues as "the file is already there" but "it's not a valid RPM file".

instead, catch the ContentTooShortError and try to delete the file

### DIFF
--- a/build_stage_repository
+++ b/build_stage_repository
@@ -3,6 +3,7 @@
 from email.utils import parsedate_to_datetime
 from subprocess import check_output, STDOUT, CalledProcessError
 from urllib.request import urlretrieve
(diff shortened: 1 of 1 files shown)


ekohl (approved):
Makes me wonder which other exceptions can be raised that leave a (partial) file in place

evgeni:
I did read https://github.com/python/cpython/blob/main/Lib/urllib/request.py before opening this PR, and from the code itself: none.
Obviously, any line could itself raise something w/o an explicit `raise`, but oh well…

PR #367, opened 2024-05-20
Add foreman 3.11

zjhuntin:


ekohl (approved):
Confirmed I can import the GPG key.

PR #372, opened 2024-05-23
Move Redmine version creation

ekohl:
The time between creating git branches and not having a Redmine version existing means the target version is wrong. The original intent of the procedure was that the release owner continued immediately with the "Other systems" part after "Branching", but this makes it obvious it needs to happen quickly after branching to minimize the window.


chris1984 (approved):
ACK

PR #373, opened 2024-06-03
add Candlepin 4.4 settings

evgeni:


releases/candlepin/4.4/settings
```diff
@@ -0,0 +1,2 @@
+FULLVERSION='4.4.5'
```
ekohl, line 1: Since e3d8ea6617a398933e7427f22bf072105c771e33 I think we no longer need to specify it, but I don't know if the candlepin flows use any of the scripts. We don't have a `procedures/candlepin` so I can't tell (nor a `procedures/pulpcore/branch.md.erb` for that matter).

ekohl (approved):
Confirmed I can import the key.

releases/candlepin/4.4/settings
evgeni: Yeah, I think we don't need it, but I also didn't want to go looking to deeply.

It's a todo for @ehelms

PR #386, opened 2024-06-27
Add GH action updates in Katello branch procedure

jeremylenz:
The link is incorrect right now, but I'm sure the reviews will fix that

### DIFF
--- a/procedures/katello/branch.md.erb
+++ b/procedures/katello/branch.md.erb
@@ -59,6 +59,8 @@
 - [ ] Push version bumps to <%= develop %>-master using <%= rel_eng_script('develop_branch_push') %>
 - [] Bump Katello related packages to their next versions
   - [] [packaging](https://github.com/theforeman/foreman-packaging/pull/8729/files)
+- [ ] Set GitHub Actions to use the new branch
+  - [ ] [React tests](https://github.com/Katello/katello/pull/11053)
 - [ ] Generate and post the release procedure, if not already posted:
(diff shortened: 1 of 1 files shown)


procedures/katello/branch.md.erb
```diff
@@ -59,6 +59,8 @@
 - [ ] Push version bumps to <%= develop %>-master using <%= rel_eng_script('develop_branch_push') %>
 - [] Bump Katello related packages to their next versions
   - [] [packaging](https://github.com/theforeman/foreman-packaging/pull/8729/files)
+- [ ] Set GitHub Actions to use the new branch
+  - [ ] [React tests](https://github.com/Katello/katello/pull/11053)
```
ekohl, line 63: I'd prefer a specific command. For example:
```suggestion
- [ ] Set GitHub Actions to use the new branch: `sed -i '/foreman_version/ s/develop/<%= foreman_version %>-stable/' .github/workflows/*.yml`
```

Note this relies on adding `foreman_version` as suggested in https://github.com/Katello/katello/pull/11053/files#r1657348705.

ekohl (changes requested):
Thanks for this. I see we didn't do this in KATELLO-4.13 so there we're testing against `develop` instead of `3.11-stable`:
https://github.com/Katello/katello/blob/cef6fb90c59ff79487e8f8ffb0a844626a9eb52d/.github/workflows/ruby.yml#L20-L27

ekohl (approved):

PR #389, opened 2024-07-15
Drop unused mash scripts

ekohl:
Since we moved to copr we no longer need these koji scripts. The workflows were modified before.

Fixes: a613d43d7079 ("Update procedures for Copr workflow")


ekohl:
I noticed I still had https://github.com/theforeman/theforeman-rel-eng/pull/214 open which made me question if this was needed. Looks like it's not.

ehelms (approved):

PR #394, opened 2024-07-30
Add Katello 4.14 rc1 and Foreman 3.12 rc1

chris1984:
I didn't add the GPG key because I am not doing the foreman release, but wanted to get this in since I made the branching doc for Katello 4.14 so I could start to work on translations.


releases/katello/4.14/settings
```diff
@@ -0,0 +1,4 @@
+FULLVERSION='4.14.0-rc1'
+FOREMAN_VERSION='3.12'
+OSES="el8 el9"
+KOJI_TAGS="katello-@VERSION@-@OS@ katello-candlepin-@VERSION@-el8"
```
ekohl, line 4: This isn't needed anymore so I've opened https://github.com/theforeman/theforeman-rel-eng/pull/395

releases/katello/4.14/settings
```diff
@@ -0,0 +1,4 @@
+FULLVERSION='4.14.0-rc1'
+FOREMAN_VERSION='3.12'
+OSES="el8 el9"
```
ekohl, line 3: I never thought about this before, but this will pretty much always match Foreman so we can drop it from Katello. https://github.com/theforeman/theforeman-rel-eng/pull/396 cleans it up.

ekohl (changes requested):
I'd prefer to wait for a GPG key to be present.

chris1984:
Closing since @pcreech is going to take it over.

PR #395, opened 2024-07-30
Drop Koji references

ekohl:
The Koji infrastructure was last used with Foreman 3.9 and after it went EOL the whole infrastructure was turned off. This cleans up the settings and scripts that were unique to it.


chris1984 (approved):
Looks fine, once this and the 396 are in I will rebase and wait until we get the GPG key

PR #414, opened 2024-08-16
Foreman 3.12 Settings file

pcreech:


releases/foreman/3.12/settings
ekohl, line 1: This is missing `OSES`

ekohl (changes requested):
I'd prefer to include https://github.com/theforeman/theforeman-rel-eng/pull/394 and https://github.com/theforeman/theforeman-rel-eng/pull/415 in here.

ekohl (approved):
Verified I can import:
```console
$ VERSION=3.12 ./import_gpg_private 
gpg: /home/ekohl/dev/theforeman-rel-eng/releases/foreman/3.12/gnupg/trustdb.gpg: trustdb created
gpg: key 4C6D3B24F1802F64: public key "Foreman Automatic Signing Key (3.12) <packages@theforeman.org>" imported
gpg: key 4C6D3B24F1802F64: secret key imported
gpg: Total number processed: 1
gpg:               imported: 1
gpg:       secret keys read: 1
gpg:   secret keys imported: 1
```

> And if we want all this in a single PR, we probably want to revisit branching procedures, as it details them in individual items.

We should because I think that's optimal.

PR #416, opened 2024-08-19
properly log errors from createrepo/modifyrepo

evgeni:


build_stage_repository
```diff
@@ -120,11 +120,21 @@ def create_modulemd(collection, version, stage_dir, foreman_version, dist):
...
+        raise
 
 
 def create_repository(repo_dir):
-    check_output(['createrepo_c', "--general-compress-type=gz", "--database" ,repo_dir], stderr=STDOUT)
+    try:
+        check_output(['createrepo_c', "--general-compress-type=gz", "--database" ,repo_dir], stderr=STDOUT)
+    except CalledProcessError as e:
+        print(e.stdout)
+        print(e.stderr)
```
ekohl, line 136: Since we already combine stderr and stdout, is this needed? Should we also write it to stderr ourselves?

build_stage_repository
```diff
@@ -120,11 +120,21 @@ def create_modulemd(collection, version, stage_dir, foreman_version, dist):
     with open(modules_yaml, 'w') as modules_file:
         yaml.dump_all(modules, modules_file, default_flow_style=False, explicit_start=True, explicit_end=True)
 
-    check_output(['modifyrepo_c', '--compress-type=gz', '--mdtype=modules', modules_yaml, f"{stage_dir}/repodata"])
+    try:
+        check_output(['modifyrepo_c', '--compress-type=gz', '--mdtype=modules', modules_yaml, f"{stage_dir}/repodata"])
```
ekohl, line 124: Should we combine stdout & stderr into one stream to maintain order?

evgeni: that's also a possibility.

build_stage_repository
ekohl: Given it's now intended to debug, I think the order is now more important.

evgeni: updated!

ekohl (approved):

PR #419, opened 2024-09-02
properly load inherited foreman settings

evgeni:
Fixes: 0ee482d44812e73fac8b535da13abfe2d711e4ed

### DIFF
--- a/settings
+++ b/settings
@@ -21,25 +21,24 @@ load_settings() {
   # Expected: FULLGPGKEY
   # Optional: FULLVERSION
   . "$RELEASEDIR/settings"
+
+  if [[ $PROJECT != foreman ]]; then
+    if [[ -z "$FOREMAN_VERSION" ]]; then
+      echo "please set FOREMAN_VERSION in $RELEASEDIR/settings"
+      exit 1
+    elif [[ "$FOREMAN_VERSION" != "none" ]]; then
+      # load foreman settings, which will contain keys etc
+      . "releases/foreman/${FOREMAN_VERSION}/settings"
+    fi
+  else
+    FOREMAN_VERSION="$VERSION"
+  fi
 }
 
 # Prepare settings file
 if [[ ! -d "$RELEASEDIR" ]] ; then
 	mkdir -p "$RELEASEDIR"
 	echo "FULLVERSION='${VERSION}.0-rc1'" > "$RELEASEDIR/settings"
-fi
-
-if [[ $PROJECT != foreman ]]; then
-  load_settings
-  if [[ -z "$FOREMAN_VERSION" ]]; then
-    echo "please set FOREMAN_VERSION in $RELEASEDIR/settings"
-    exit 1
-  elif [[ "$FOREMAN_VERSION" != "none" ]]; then
-    # load foreman settings, which will contain keys etc
-    . "releases/foreman/${FOREMAN_VERSION}/settings"
-  fi
-else
-  FOREMAN_VERSION="$VERSION"
 fi
 
 SIGNER="packages@theforeman.org"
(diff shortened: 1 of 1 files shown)


settings
```diff
@@ -21,6 +21,18 @@ load_settings() {
   # Expected: FULLGPGKEY
   # Optional: FULLVERSION
   . "$RELEASEDIR/settings"
+
+  if [[ $PROJECT != foreman ]]; then
```
ekohl, line 25: Why do we need to do this before creating a dummy settings file?

settings
```diff
@@ -70,6 +67,8 @@ GPG_EXPIRE="1y"
 
 load_settings
 
+PASS_NAME_GPG="theforeman/releases/foreman/$FOREMAN_VERSION-gpg"
+PASS_NAME_KEY="theforeman/releases/foreman/$FOREMAN_VERSION-key"
```
ekohl, line 71: This overrides the settings, which means you can't override it anymore as we do:
```console
releases/pulpcore/settings
2:PASS_NAME_GPG="theforeman/releases/pulpcore/$VERSION-gpg"
3:PASS_NAME_KEY="theforeman/releases/pulpcore/$VERSION-key"

releases/foreman-debian/settings
1:PASS_NAME_GPG="theforeman/releases/foreman/debian-$VERSION-gpg"
2:PASS_NAME_KEY="theforeman/releases/foreman/debian-$VERSION-key"

releases/candlepin/settings
1:PASS_NAME_GPG="theforeman/releases/candlepin/$VERSION-gpg"
2:PASS_NAME_KEY="theforeman/releases/candlepin/$VERSION-key"
```

It's also not like we can use `$PROJECT` like this:
```suggestion
PASS_NAME_GPG="theforeman/releases/$PROJECT/$VERSION-gpg"
PASS_NAME_KEY="theforeman/releases/$PROJECT/$VERSION-key"
```

What I'm leaning to is move these definitions into `releases/foreman/settings` (and not have any default) but I don't think we load that now when inheriting settings.

Alternatively you can use:
```suggestion
PASS_NAME_GPG="${PASS_NAME_GPG:-theforeman/releases/foreman/$FOREMAN_VERSION-gpg}"
PASS_NAME_KEY="${PASS_NAME_KEY:-theforeman/releases/foreman/$FOREMAN_VERSION-key}"
```

ekohl (changes requested):

settings
evgeni: this is not before -- it's a function :)

settings
evgeni: dang.

so the problem I was trying to fix is:
When signing Katello, `$OSES` ends up `''` as it first loaded `katello/4.14/settings`, then `foreman/3.12/settings` and then overwrote `OSES` with the value in `settings`.
To fix that, I moved the "load inherited foreman stuff" logic into `load_settings`, but that that meant that `FOREMAN_VERSION` is not yet populated when the old location of `PASS_NAME_*` was used, so I moved that too.

settings
ekohl: Right, my bad.

settings
ekohl: I added a commit that I think addresses the concerns. It's not pretty, but I think it'll work.

settings
```diff
@@ -21,6 +21,17 @@ load_settings() {
...
   . "$RELEASEDIR/settings"
+
+  if [[ $PROJECT != foreman ]]; then
+    if [[ -z "$FOREMAN_VERSION" ]]; then
+      echo "please set FOREMAN_VERSION in $RELEASEDIR/settings"
+      exit 1
+    elif [[ "$FOREMAN_VERSION" != "none" ]]; then
+      # load foreman settings, which will contain keys etc
+      . releases/foreman/settings
+      . "releases/foreman/${FOREMAN_VERSION}/settings"
```
evgeni, line 32: mhh, no, this overwrites OSes that have been set in `client/settings` before. damn.

PR #420, opened 2024-09-03
Update upload stage RPMs to new host

ehelms:
This would accompany our change to where RPMs are hosted -- https://github.com/theforeman/foreman-infra/pull/2118

For staging RPMs, the structure remains the same and nothing but the host will have to change when we are ready.


upload_stage_rpms
```diff
@@ -2,7 +2,11 @@
 
 . settings
 
-USER='yumrepostage'
-HOST='web01.osuosl.theforeman.org'
+USER='rpmrepostage'
+HOST='rpm.theforeman.org'
```
ekohl, line 6: You need to list the host explicitly. `rpm.theforeman.org` will be the CDN and you can't SSH into that.

upload_stage_rpms
```diff
@@ -2,7 +2,11 @@
...
 
-rsync --checksum --times --perms --recursive --links --verbose --partial --one-file-system --delete-after "tmp/$PROJECT/$VERSION/" "$USER@$HOST:rsync_cache/$PROJECT/$VERSION/"
+if [[ "$PROJECT" == "pulpcore" || "$PROJECT" == "candlepin" ]];then
+  rsync --checksum --times --perms --recursive --links --verbose --partial --one-file-system --delete-after "tmp/$PROJECT/$VERSION/" "$USER@$HOST:rsync_cache/$PROJECT/$VERSION/"
+else
+  rsync --checksum --times --perms --recursive --links --verbose --partial --one-file-system --delete-after "tmp/$PROJECT/$VERSION/" "$USER@$HOST:rsync_cache/foreman/$VERSION/$PROJECT/"
```
ekohl, line 11: I think this needs to be:
```suggestion
  rsync --checksum --times --perms --recursive --links --verbose --partial --one-file-system --delete-after "tmp/$PROJECT/$VERSION/" "$USER@$HOST:rsync_cache/foreman/$FOREMAN_VERSION/$PROJECT/"
```

publish_release_rpms
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
+
+. settings
+
+USER='rpmrepo'
+HOST='rpm.theforeman.org'
```
ekohl, line 6: You need to list the host explicitly. `rpm.theforeman.org` will be the CDN and you can't SSH into that.

ekohl (changes requested):

PR #426, opened 2024-09-09
drop signing scripts that were used with koji

evgeni:


pcreech (approved):

ekohl:
I thought I had taken care of everything in https://github.com/theforeman/theforeman-rel-eng/pull/395 but clearly it was incomplete.

PR #436, opened 2024-10-24
Foreman 3.12 GA config changes

pcreech:


releases/katello/4.14/settings
```diff
@@ -1,2 +1,3 @@
-FULLVERSION='4.14.0-rc1'
+FULLVERSION='4.14.0'
 FOREMAN_VERSION='3.12'
+OSES='el8 el9'
```
ekohl, line 3: This should be inherited from Foreman itself now.

ekohl (changes requested):
There are merge conflicts now. Could you rebase?

releases/katello/4.14/settings
ekohl:
```suggestion
```

Issue #417, opened 2024-08-20
Update to the release procedures:  Whomever creates foreman settings file for branching should also create katello and client settings


pcreech:
Whomever generates the foreman settings file for a given minor version should also generate the katello and client settings files

pcreech:
Updated based on feedback from https://github.com/theforeman/theforeman-rel-eng/pull/414#pullrequestreview-2248035400

PR #437, opened 2024-10-28
Add Katello 4.15

sambible:


releases/katello/4.15/settings
```diff
@@ -0,0 +1,3 @@
+FULLVERSION='4.15.0'
```
chris1984, line 1: @ekohl should this be rc1 or is it fine to keep at 4.15.0?

releases/katello/4.15/settings
```diff
@@ -0,0 +1,3 @@
...
+FOREMAN_VERSION='3.13'
+OSES="el8 el9"
```
ekohl, line 3: Since https://github.com/theforeman/theforeman-rel-eng/commit/0ee482d44812e73fac8b535da13abfe2d711e4ed it's better to leave it out since it's inherited from Foreman's config.
```suggestion
```

releases/katello/4.15/settings
ekohl, line 1: We start with rc1
```suggestion
FULLVERSION='4.15.0-rc1'
```

ekohl (changes requested):

chris1984 (approved):
LGTM @ekohl have all your comments been addressed?

ekohl (commented):
The contents are OK, but I have a bit of an issue that this right now depends on Foreman's config (how else is it going to inherit that?) which is still missing. We also ran into this last time, but I really think this should be done together with Foreman 3.13's config. 

Right now we have this procedure for Foreman:

https://github.com/theforeman/theforeman-rel-eng/blob/6587a92d786f37ac26a3b5d602a455ccadf75f2d/procedures/foreman/branch.md.erb#L56-L65

And then for Katello:

https://github.com/theforeman/theforeman-rel-eng/blob/6587a92d786f37ac26a3b5d602a455ccadf75f2d/procedures/katello/branch.md.erb#L35

Could you (plural, so including @chris1984) take a look at the procedures to see how we can prevent this from again being an issue with Katello 4.16?

On the solution: we already let the release engineer create the Jenkins config for Katello:
https://github.com/theforeman/theforeman-rel-eng/blob/6587a92d786f37ac26a3b5d602a455ccadf75f2d/procedures/foreman/branch.md.erb#L66

To me it feels natural to also instruct the release owner to prepare Katello's rel-eng settings (which implies dropping it from Katello's procedure), but making the dependency explicit in Katello's procedure is also fine by me.

I'd appreciate @pcreech to weigh in as well since he's often the release engineer doing the Foreman side.

chris1984:
> The contents are OK, but I have a bit of an issue that this right now depends on Foreman's config (how else is it going to inherit that?) which is still missing. We also ran into this last time, but I really think this should be done together with Foreman 3.13's config.
> 
> Right now we have this procedure for Foreman:
> 
> https://github.com/theforeman/theforeman-rel-eng/blob/6587a92d786f37ac26a3b5d602a455ccadf75f2d/procedures/foreman/branch.md.erb#L56-L65
> 
> And then for Katello:
> 
> https://github.com/theforeman/theforeman-rel-eng/blob/6587a92d786f37ac26a3b5d602a455ccadf75f2d/procedures/katello/branch.md.erb#L35
> 
> Could you (plural, so including @chris1984) take a look at the procedures to see how we can prevent this from again being an issue with Katello 4.16?
> 
> On the solution: we already let the release engineer create the Jenkins config for Katello:
> 
> https://github.com/theforeman/theforeman-rel-eng/blob/6587a92d786f37ac26a3b5d602a455ccadf75f2d/procedures/foreman/branch.md.erb#L66
> 
> To me it feels natural to also instruct the release owner to prepare Katello's rel-eng settings (which implies dropping it from Katello's procedure), but making the dependency explicit in Katello's procedure is also fine by me.
> 
> I'd appreciate @pcreech to weigh in as well since he's often the release engineer doing the Foreman side.

I went through the branching procedure and I agree it makes sense. Foreman should be branching before Katello and the release engineer would if all things are going smooth have made the Foreman config and so it makes sense while working on the Foreman config, create the Katello config. I also while going through the release procedure did find a 404 with the doc link https://github.com/theforeman/theforeman-rel-eng/pull/439

sambible:
https://github.com/theforeman/theforeman-rel-eng/issues/417

sambible:
https://github.com/theforeman/theforeman-rel-eng/issues/445 - an issue that captures the doc update request, and links back to here.

chris1984:
@ekohl I had Sam create this issue - https://github.com/theforeman/theforeman-rel-eng/issues/445 so we don't hit this in Katello 4.16. Are you fine with that and we can get this in?

ekohl:
> The only question that comes up is if you are doing a Katello z release where Foreman is not involved, so maybe have a step in the release procedure doc saying if this is a z release then update the config, so the Katello owner is only updating instead of creating the config.

That's the release procedure and I think that should remain unchanged. I was only talking about the branching procedure.

ekohl (approved):

PR #442, opened 2024-11-05
Procedure updates post-3.13 branching

pcreech:


procedures/foreman/branch.md.erb
```diff
@@ -59,6 +59,7 @@
   - [ ] <%= rel_eng_script('sign_gpg') %>
   - [ ] <%= rel_eng_script('upload_gpg') %>
 - Publish the key
+  - [ ] Sync gopass `gopass sync`
```
ekohl, line 62: We have the `export_gpg_private` which calls `gopass insert`. I have a vague recollection that it also syncs it normally. I wonder if that somehow failed. If not, perhaps we should enhance that script to call `gopass sync`?

https://woile.github.io/gopass-cheat-sheet/ tells me there's also `gopass sync --store $store` so 

On a related note, `import_gpg_private` could probably also call `gopass sync`.

In https://github.com/theforeman/theforeman-rel-eng/pull/443 I implemented that.

pcreech: Cool.  I'll keep this draft PR open though to ensure I capture any of these i come across.

PR #447, opened 2024-11-12
Pin Foreman's CI to stable branches on branching

### LINKED ISSUE #583: During branching, the version of katello used in tests should be pinned
To prevent tests on a stable branch of foreman from running with latest katello.

https://github.com/theforeman/foreman/pull/10999#issuecomment-4543872712

ekohl:
I'm not entirely happy with this, because it makes it interactive. I don't see another way, though it should be noted we may have more cases where we need to know the Katello version when branching. For example, jenkins-jobs but maybe more.

### DIFF
--- a/branch_project
+++ b/branch_project
@@ -14,6 +14,20 @@ for repo in $(./ensure_clean_git_checkouts "$@") ; do
 			bundle exec librarian-puppet install
 			git add .gitignore Puppetfile*
 			git commit --message "Pin Puppet modules for ${VERSION}"
+		elif [[ $repo == foreman ]] ; then
+			# TODO: We don't know the Katello version at this point
+			read -p "Katello version: " -r KATELLO
+			if [[ -z $KATELLO ]] ; then
+				echo "No Katello version given" >&2
+				exit 1
+			fi
+			sed -i "/plugin_repository/a\      plugin_version: KATELLO-$KATELLO" .github/workflows/foreman.yml
+			git add .github/workflows/foreman.yml
+
+			sed -i "s/nightly/$VERSION/" .packit.yaml
+			git add .packit.yaml
+
+			git commit --message "Adjust CI to use $VERSION stable branches"
 		elif [[ $repo == candlepin-packaging ]] ; then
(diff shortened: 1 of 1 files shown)


evgeni:
Should the Katello part be a step in the Katello branching?
Or could we fetch/parse https://github.com/Katello/katello/blob/master/lib/katello/version.rb on the fly? (Assuming Katello is never branched before Foreman)

(Not saying those are good ideas, but wanted to throw them regardless)

ekohl:
I was thinking about introducing `KATELLO_VERSION` in settings, similar to how we have `FOREMAN_VERSION`:
https://github.com/theforeman/theforeman-rel-eng/blob/434242abaf30716372f1875cf9be6f0acedca979/releases/katello/4.15/settings#L2

Might not be pretty, but feels like the most reliable thing.

> For example, jenkins-jobs but maybe more.

Today we have:

https://github.com/theforeman/theforeman-rel-eng/blob/434242abaf30716372f1875cf9be6f0acedca979/procedures/foreman/branch.md.erb#L66
https://github.com/theforeman/theforeman-rel-eng/blob/434242abaf30716372f1875cf9be6f0acedca979/procedures/foreman/branch.md.erb#L77

So we already need it in 2 places. Isn't 3 the magic number to properly fix it?

Also, we can probably use the same approach to update https://github.com/Katello/katello/blob/d07494f4dba5f32f4673eb78ada8e36bacfa985d/.github/workflows/ruby.yml#L30 in an automated way.

ekohl:
> Hmm, can't we enumerate over all `releases/katello/*/settings` and find the one that has the right `FOREMAN_VERSION`?

I thought about that, but it felt a bit magical

> Otherwise it seems like we're pointing in circles

I wasn't thinking about loading the Katello settings, but yes, it's a bit redundant.

Thinking higher level, we could restructure things to have a sub-project. So `PROJECT=foreman SUBPROJECT=katello` or `PROJECT=foreman SUBPROJECT=client`. Not sure I want to go in that direction now, but it could simplify our inheritance loading by swapping it around.

ogajduse:
Superseded by #607

PR #472, opened 2025-02-26
Fix small typos in Katello branching and release processes; add helpful clarifications

qcjames53:
Fix small typos in Katello branching and release processes; add helpful clarifications.

I was routinely running into small discrepancies between the Foreman and Katello branching and release process scripts. I have made several changes to fix small issues in the processes and to clarify some sticking points:

- The branching instructions now name the first release candidate as `VERSION.rc1` instead of `VERSION-rc1`. Katello differs from Foreman in that we use dots for release candidates.
- Cherry pick instructions add the necessary PROJECT and VERSION bash variables.
- A few `./tools` are now `./tools.rb`.
- Developers are told they can skip the foreman-documentation section after rc1 if there are no new headline features or contributors.
- Added a shell command to generate an alphabetized contributors list.
- Changelog instructions add the necessary PROJECT and VERSION bash variables.
- Final release tagging instructions lead to a detached head state so the `git push` command has been modified to allow pushes to upstream from this state.
- Katello gem pushes require 2FA; instructions have been modified to reflect this.
- Added instructions for how to generate release instructions for the next release.


procedures/foreman/release.md.erb
```diff
@@ -99,7 +99,7 @@ Note: If for some reason there was an issue with the tarballs that required uplo
 <% if is_first_rc || (!is_rc && !is_first_ga) -%>
   - Add `topics/foreman-<%= version %>.adoc` to `guides/doc-Release_Notes/master.adoc`: `sed -i '/x.y.z releases here/a include::topics/foreman-<%= version %>.adoc[leveloffset=+1]' guides/doc-Release_Notes/master.adoc`
 <% end -%>
-  - Make sure [foreman-contributors.adoc](https://github.com/theforeman/foreman-documentation/blob/<%= short_version %>/guides/doc-Release_Notes/topics/foreman-contributors.adoc) is updated
+  - Update [foreman-contributors.adoc](https://github.com/theforeman/foreman-documentation/blob/<%= short_version %>/guides/doc-Release_Notes/topics/foreman-contributors.adoc) with the list of contributors using `git log --since="<YYYY-MM-DD of previous major release branching>" --format="%an" | sort | uniq`. Manually clean up any erroneous entries.
```
ekohl, line 102: We already have https://github.com/theforeman/theforeman.org/blob/gh-pages/scripts/committers.rb that should be used.

That repo also contains a longer [.mailmap](https://github.com/theforeman/theforeman.org/blob/gh-pages/.mailmap) to map bad entries to normal users. To use those, it uses `%aN` instead. It's also based on git tags, not dates.

procedures/katello/release.md.erb
```diff
@@ -32,7 +32,7 @@
 <% unless is_rc && !full_version.end_with?('1') -%>
   - Add `topics/katello-<%= version %>.adoc` to `guides/doc-Release_Notes/master.adoc`: `sed '/x.y.z releases here/a include::topics/katello-<%= version %>.adoc[leveloffset=+1]' master.adoc`
 <% end -%>
-  - Make sure [katello-contributors.adoc](https://github.com/theforeman/foreman-documentation/blob/<%= foreman_version %>/guides/doc-Release_Notes/topics/katello-contributors.adoc) is updated
+  - Update [katello-contributors.adoc](https://github.com/theforeman/foreman-documentation/blob/<%= foreman_version %>/guides/doc-Release_Notes/topics/katello-contributors.adoc) with the list of contributors using `git log --since="<YYYY-MM-DD of previous major release branching>" --format="%an" | sort | uniq`. Manually clean up any erroneous entries.
```
ekohl, line 35: Any chance we could standardize things by using the same script or reuse as much as possible?

procedures/foreman/release.md.erb
qcjames53: Wow that's a helpful script. I'll adjust this PR in a bit to reflect this info. Thanks, Ewoud!

qcjames53: @ekohl is there a way to run this script with the Katello repo? Is `katello` an option for `--with-repo`? I'd be happy to add it to the release process but unfortunately did not have enough time to figure this out.

qcjames53:
I kept track of every minor issue with the Katello branching and release processes and added the tweaks to this PR. 

I would be happy to clone the parts that make sense to the Foreman script in the not too distant future but I unfortunately don't have the time for that today.

procedures/foreman/release.md.erb
ianballou: It looks like this is not implemented for Katello yet. I tried adding Katello to the list but it causes git to ask for a username and password. Seems it needs more looking into.

procedures/katello/release.md.erb
```diff
@@ -88,3 +88,10 @@
 - [ ] Create and upload [release specific developer stable box](https://github.com/theforeman/forklift/blob/master/packer/README.md#creating-version-specific-katello-devel-boxes)
 - [ ] Remove references to the oldest Katello version from [testKatello.groovy](https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/test/testKatello.groovy) and [katello-pipelines.yml](https://github.com/theforeman/jenkins-jobs/blob/master/centos.org/jobs/katello-pipelines.yml).
 <% end -%>
+
+# When ready to release next Katello <%= short_version %> release candidate or minor version
+### Release Owner
```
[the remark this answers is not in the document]
qcjames53: Nice catch!

qcjames53:
Thanks for the review, Ian. The small typo should now be fixed.

procedures/foreman/release.md.erb
qcjames53: I'm happy to leave this as the ` | sort | uniq` solution. It's not too difficult at all to clean up the contributors' names manually.

ianballou (approved):
I'd like to get this through since it has good updates that the next owner will need. We can update the instructions if the contributor-fetching script adds Katello support.

ekohl (commented):
Review from my phone. Likely that I missed things and on my desktop I would have added more links

procedures/katello/branch.md.erb
```diff
@@ -33,6 +33,7 @@
 ## Release Owner
 
 - [ ] Commit the [settings file](https://github.com/theforeman/theforeman-rel-eng/blob/master/releases/katello/<%= release %>/settings) to the `theforeman-rel-eng` repository. See an example [here](https://github.com/theforeman/theforeman-rel-eng/commit/29b6b92e82de25c7a0e32fb9ca36468d8df0f6a1)
+  - [ ] Ensure that the version string uses `VERSION.rc1` instead of `VERSION-rc1`. Katello's syntax differs from Foreman.
```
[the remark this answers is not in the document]
qcjames53: Unfortunately this is correct. We already had subpoint check boxes in other places in this document so I'm personally okay adding a few more. In cases like this it's not really mandatory but it's still nice to let folks keep track of their progress imho.

procedures/katello/branch.md.erb
```diff
@@ -62,7 +63,7 @@
 - [ ] Set GitHub Actions to use the new branch: `sed -i '/foreman_version/ s/develop/<%= foreman_version %>-stable/' .github/workflows/*.yml`
 - [ ] Generate and post the release procedure, if not already posted:
   - [ ] Use <%= rel_eng_script('procedure_release') %>: `PROJECT=katello VERSION=<%= release %> ./procedure_release "<%= target_date %>" "<%= owner %>" "<%= engineer %>"`
-  - [ ] Post the output in [Development/Releases](https://community.theforeman.org/c/development/releases/20) using `Katello <%= release %>.0-rc1 release process`
+  - [ ] Post the output in [Development/Releases](https://community.theforeman.org/c/development/releases/20) using `Katello <%= release %>.0.rc1 release process`
```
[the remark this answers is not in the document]
qcjames53: Unfortunately the branching process doesn't have access. I can add it if you would like but I really don't think it's necessary. I don't believe we'll be branching for anything except `VERSION.0.rc1`.

procedures/katello/release.md.erb
```diff
@@ -14,42 +14,42 @@
...
 <% unless is_rc && !full_version.end_with?('1') -%>
-  - Add `topics/katello-<%= version %>.adoc` to `guides/doc-Release_Notes/master.adoc`: `sed '/x.y.z releases here/a include::topics/katello-<%= version %>.adoc[leveloffset=+1]' master.adoc`
+  - [ ] Add `topics/katello-<%= version %>.adoc` to `guides/doc-Release_Notes/master.adoc`: `sed '/x.y.z releases here/a include::topics/katello-<%= version %>.adoc[leveloffset=+1]' master.adoc`
 <% end -%>
-  - Make sure [katello-contributors.adoc](https://github.com/theforeman/foreman-documentation/blob/<%= foreman_version %>/guides/doc-Release_Notes/topics/katello-contributors.adoc) is updated
-  - Submit this as a PR
+  - [ ] Update [katello-contributors.adoc](https://github.com/theforeman/foreman-documentation/blob/<%= foreman_version %>/guides/doc-Release_Notes/topics/katello-contributors.adoc) with the list of contributors using `git log --since="<YYYY-MM-DD of previous major release branching>" --format="%an" | sort | uniq`. Manually clean up any erroneous entries.
```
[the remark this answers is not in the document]
qcjames53: I like that idea a lot. I'll keep this as date-based for now and we can switch over if/when we make the change down the road.

procedures/katello/release.md.erb
```diff
@@ -88,3 +88,10 @@
 - [ ] Create and upload [release specific developer stable box](https://github.com/theforeman/forklift/blob/master/packer/README.md#creating-version-specific-katello-devel-boxes)
 - [ ] Remove references to the oldest Katello version from [testKatello.groovy](https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/test/testKatello.groovy) and [katello-pipelines.yml](https://github.com/theforeman/jenkins-jobs/blob/master/centos.org/jobs/katello-pipelines.yml).
 <% end -%>
+
+# When ready to release next Katello <%= short_version %> release candidate or minor version
```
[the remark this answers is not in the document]
qcjames53: It doesn't look like we have this for rc2, just for branching (generating rc1). I added this for all releases since I was finding it difficult to remember the command to generate release procedures for the .1+ releases. I felt that there was no way to know how rigid our release schedule would be in the future so I left it up to the developer.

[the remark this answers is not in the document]
qcjames53: I'd prefer to leave this in for all releases due to the reasoning above. What are your thoughts?

qcjames53:
@ekohl As requested, I added the transifex section to non-rc1 releases so our translations are always updated. It's a copy-paste job but I can always turn it into a method if you'd prefer to avoid the instruction duplication.

procedures/katello/release.md.erb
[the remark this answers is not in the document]
ekohl: I experimented with it in https://github.com/theforeman/foreman/pull/10512. There are some practical concerns so just relating this back.

procedures/katello/release.md.erb
```diff
@@ -14,42 +14,42 @@
...
+  - [ ] Update/add the CHANGELOG.md file: `GITHUB_ACCESS_TOKEN=<secret> PROJECT=katello VERSION=<%= short_version %> ./tools.rb changelog configs/katello/<%= short_version %>.yaml`
   - [ ] Commit: `git commit -m "Release <%= full_version %>"`
   - [ ] Ensure that the commit above is the _last_ commit and there are no commits after it. This is the commit that will get tagged. (Rearrange commits with `git rebase -i` if needed.)
 - [ ] Once the PR is merged, perform the following in the Katello release branch (the real one, not your fork):
   - [ ] Create upstream remote: `git remote add upstream https://github.com/Katello/katello.git`
   - [ ] Fetch upstream remote: `git fetch upstream`
   - [ ] Checkout upstream release branch: `git checkout upstream/KATELLO-<%= short_version %>`
   - [ ] Tag: `git tag -s -m "Release <%= full_version %>" <%= full_version %>`
-  - [ ] Push: `git push --follow-tags` (Must be pushed directly to the release branch, as pull request merges will not preserve tags.)
+  - [ ] Push: `git push --follow-tags upstream HEAD:KATELLO-<%= short_version %>` (Must be pushed directly to the release branch, as pull request merges will not preserve tags.)
```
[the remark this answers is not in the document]
ekohl: 457f7f54b08c53719b8b85b348f9abff881e52f1 intended to make it work, but looks like the process is different enough because Katello submits a PR before and assumes the commit is already made where Foreman has the script make that commit.

Once we resolve that, you should be able to copy the exact same instructions as Foreman:
https://github.com/theforeman/theforeman-rel-eng/blob/de3b841f73facf6d1a61db1cf79fb2125085345c/procedures/foreman/release.md.erb#L35-L37

Then the steps to generate the `.mo` files and call `gem build` etc would be equivalent to Foreman's `release_tarballs` step (since the tarballs are the release artifacts for Foreman). That should be kept in for now.

We can also automate gem releases. We already have that for other plugins after a tag is pushed. See https://community.theforeman.org/t/use-trusted-publishers-to-release-gems/42009 for how we've done that. Only exception is that the `.mo` file building isn't part of it, but we can move that to RPM packaging instead.

Having said all of that: it would rely on updating the settings to match. Right now the version isn't updated since RC1:
https://github.com/theforeman/theforeman-rel-eng/blob/de3b841f73facf6d1a61db1cf79fb2125085345c/releases/katello/4.16/settings#L1

I'd prefer to capture this in an issue upstream and if you want you can point a JIRA issue at it. Thoughts?

procedures/katello/branch.md.erb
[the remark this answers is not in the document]
ekohl: We have some code to initialize the `settings` file:
https://github.com/theforeman/theforeman-rel-eng/blob/de3b841f73facf6d1a61db1cf79fb2125085345c/settings#L26-L30

The idea is that if you run any command with a version that doesn't exist it will initialize it for you.

Perhaps it's an idea to enhance that to use the correct format for Katello and also write down `FOREMAN_VERSION=` which is required? Then introduce since you already run `procedure_branch` it will already have the correct content.

https://github.com/theforeman/theforeman-rel-eng/pull/482 implements that and IMHO makes this line redundant.

procedures/katello/release.md.erb
```diff
@@ -14,42 +14,54 @@
 - [ ] Do Cherry-picks: Clone [tool_belt](https://github.com/theforeman/tool_belt)
   - [ ] If any minor versions of Katello have been added to Redmine since the last cherry-pick, make sure to include them in `prior_releases` in configs/katello/<%= short_version %>.yaml
   - [ ] Run `./tools.rb setup-environment configs/katello/<%= short_version %>.yaml`
-  - [ ] Run `GITHUB_ACCESS_TOKEN=<secret> ./tools.rb cherry-picks --version <%= full_version %> configs/katello/<%= short_version %>.yaml`
+  - [ ] Run `GITHUB_ACCESS_TOKEN=<secret> PROJECT=katello VERSION=<%= short_version %> ./tools.rb cherry-picks configs/katello/<%= short_version %>.yaml`
```
ekohl, line 17: This doesn't look correct because tool_belt doesn't respect `PROJECT` and `VERSION`.

procedures/katello/release.md.erb
```diff
@@ -14,42 +14,54 @@
...
   - [ ] Open a PR in Katello release branch.  Make sure the PR name starts with [CP] to prevent our automations from adding it to Redmine issues.
   - [ ] Using `git cherry-pick -x` as needed, verify tickets in the cherry_picks_<%= full_version %> file are accounted for, or additionally cherry-pick them.  Recommended: Do this in tool_belt's checkout of Katello, in `repos/katello/<%= full_version %>/katello`. This way when you run cherry-picks again, tool_belt will be aware of any picks already completed.
   - [ ] For any cherry-picks that are not needed (including Redmine trackers) you can add them to the `:ignores:` section of `tool_belt` in `configs/katello/<%= short_version %>.yaml`
+- [ ] Update Katello Transifex translations:
+  - [ ] Create a Transifex account and join the Foreman team
+  - [ ] Spin up a Foreman and Katello installation
+  - [ ] [Configure the Transifex client](https://projects.theforeman.org/projects/foreman/wiki/Translating#How-to-pull-translations)
+  - [ ] [Install grunt](https://github.com/Katello/katello/blob/master/engines/bastion_katello/README.md)
+  - [ ] `grunt i18n:extract` in the katello/engines/bastion_katello directory
```
ekohl, line 26: Why doesn't the `Makefile` do this?

procedures/katello/release.md.erb
```diff
@@ -14,42 +14,54 @@
...
+  - [ ] `tx pull --minimum-perc 50 --all` in the katello directory
```
ekohl, line 27: Since https://github.com/Katello/katello/commit/2607abb55b6154c8f2763dc444b418ce48d3eb06 `Makefile` does this when running `tx-pull`
```suggestion
```

procedures/katello/release.md.erb
```diff
@@ -14,42 +14,54 @@
...
+  - [ ] `make -C locale tx-pull` in the katello directory
+  - [ ] `grunt i18n:compile` in the katello/engines/bastion_katello directory
+  - [ ] `bundle exec rake plugin:gettext[katello]` in the foreman directory
+  - [ ] `bundle exec rake plugin:po_to_json[katello]` in the foreman directory
```
ekohl, line 31: If you run `tx-update` then it automatically runs `tx-pull` and tells you the next steps. I'd prefer to have it in code as much as possible.

procedures/katello/release.md.erb
```diff
@@ -14,42 +14,54 @@
...
   - [ ] For any cherry-picks that are not needed (including Redmine trackers) you can add them to the `:ignores:` section of `tool_belt` in `configs/katello/<%= short_version %>.yaml`
+- [ ] Update Katello Transifex translations:
+  - [ ] Create a Transifex account and join the Foreman team
+  - [ ] Spin up a Foreman and Katello installation
+  - [ ] [Configure the Transifex client](https://projects.theforeman.org/projects/foreman/wiki/Translating#How-to-pull-translations)
+  - [ ] [Install grunt](https://github.com/Katello/katello/blob/master/engines/bastion_katello/README.md)
+  - [ ] `grunt i18n:extract` in the katello/engines/bastion_katello directory
+  - [ ] `tx pull --minimum-perc 50 --all` in the katello directory
+  - [ ] `make -C locale tx-pull` in the katello directory
+  - [ ] `grunt i18n:compile` in the katello/engines/bastion_katello directory
```
ekohl, line 29: Why doesn't the `Makefile` do this?

procedures/katello/release.md.erb
```diff
@@ -14,42 +14,54 @@
...
+  - [ ] `bundle exec rake plugin:gettext[katello]` in the foreman directory
+  - [ ] `bundle exec rake plugin:po_to_json[katello]` in the foreman directory
+  - [ ] `make -C locale commit-translation-files` in the katello directory
```
ekohl, line 32: Looks like https://github.com/Katello/katello/commit/2607abb55b6154c8f2763dc444b418ce48d3eb06 removed `-commit-translation-files` in an effort to become standardized with all other plugins. Those use `mo-files` though arguably that's misnamed now. At least `tx-update` tells you what to run.

Now I'm unsure what to do here. Should we fix up `Makefile` so the procedure becomes consistent again between versions or should we move forward, accepting that for 4.16.1 it'll be incorrect?

ekohl (commented):
Thanks for adding the translation steps. I have some questions and would appreciate your input.

procedures/katello/branch.md.erb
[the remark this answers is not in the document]
qcjames53: Totally agree with you here. #482 solves the issue entirely. I'll remove the line.

procedures/katello/release.md.erb
```diff
@@ -14,42 +14,54 @@
...
-  - Submit this as a PR
-
+  - [ ] Update [katello-contributors.adoc](https://github.com/theforeman/foreman-documentation/blob/<%= foreman_version %>/guides/doc-Release_Notes/topics/katello-contributors.adoc) with the list of contributors using `git log --since="<YYYY-MM-DD of previous major release branching>" --format="%an" | sort | uniq`. Manually clean up any erroneous entries.
+  - [ ] Submit this as a PR
 - [ ] Open a PR (or use cherry-pick PR) against the release branch which updates `lib/katello/version.rb` to <%= full_version %>:
   - [ ] `git pull` to make sure you have the latest changes
   - [ ] `sed '/VERSION/ s/".\+"/"<%= full_version %>"/' lib/katello/version.rb`
-  - [ ] Update/add the CHANGELOG.md file: `GITHUB_ACCESS_TOKEN=<secret> ./tools changelog --version <%= full_version %> configs/katello/<%= short_version %>.yaml`
+  - [ ] Update/add the CHANGELOG.md file: `GITHUB_ACCESS_TOKEN=<secret> PROJECT=katello VERSION=<%= short_version %> ./tools.rb changelog configs/katello/<%= short_version %>.yaml`
```
qcjames53, line 53: The original syntax seems to work fine for this line as well. I'm reverting the changes but keeping the '.rb'.

procedures/katello/release.md.erb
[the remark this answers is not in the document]
qcjames53: I think that sounds like a plan. I can create an issue/JIRA for this.

procedures/katello/release.md.erb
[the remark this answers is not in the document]
qcjames53: I agree that the readme's release instructions should cover this issue. I hadn't noticed that existed. I have pulled out the new section.

qcjames53:
This should be good to go now. I can create the translations PR shortly.

PR #475, opened 2025-03-11
Update settings to 3.14.0 GA

ColeHiggins2:


ekohl (approved):
I think all code is in and we can reasonably say we won't do an RC3.

PR #479, opened 2025-04-07
Replace 'edit -c' by 'insert'

lhellebr:
I'm not sure why but 'edit --create' failed with "Error: no results found". Using insert worked for me.


lhellebr:
I'm using the Fedora 41 pre-packaged version.

Opening an editor works for me, this is some different issue I don't yet understand... but this change fixed it for me.

ekohl:
I've merged https://github.com/theforeman/theforeman-rel-eng/pull/484 instead because I think it's cleaner.

PR #487, opened 2025-05-07
Only generate settings for foreman and katello projects

evgeni:
Fixes: 023c84c8fb49d1eec7c1079d268c95330636aad5

### DIFF
--- a/settings
+++ b/settings
@@ -20,7 +20,9 @@ load_settings() {
   # Load settings per release
   # Expected: FULLGPGKEY
   # Optional: FULLVERSION
-  . "$RELEASEDIR/settings"
+  if [[ -f "${RELEASEDIR}/settings" ]]; then
+    . "$RELEASEDIR/settings"
+  fi
 }
 
 # Prepare settings file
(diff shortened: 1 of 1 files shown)


settings
```diff
@@ -20,18 +20,20 @@ load_settings() {
...
 if [[ ! -d "$RELEASEDIR" ]] ; then
 	mkdir -p "$RELEASEDIR"
 	if [[ $PROJECT == katello ]] ; then
 		echo "FULLVERSION='${VERSION}.0.rc1'" > "$RELEASEDIR/settings"
-	else
+	elif [[ $PROJECT == foreman ]]; then
 		echo "FULLVERSION='${VERSION}.0-rc1'" > "$RELEASEDIR/settings"
 	fi
-	if [[ $PROJECT != foreman ]] && [[ $FOREMAN_VERSION != "none" ]] ; then
+	if [[ $PROJECT != foreman ]] && [[ $FOREMAN_VERSION != "none" ]] && [[ -n "$FOREMAN_VERSION" ]] ; then
```
ekohl, line 36: I added this so there was an explicit empty value as well because it is expected to be set.

settings
```diff
@@ -20,18 +20,20 @@ load_settings() {
   # Load settings per release
   # Expected: FULLGPGKEY
```
ekohl, line 21: Technically this also isn't always required. Otherwise we should add this to the "template".

ekohl (commented):
How about we move the `Prepare settings file` step into `load_settings()`? After `load project specific settings first` (so we load `FOREMAN_VERSION="none"`) and before `Load settings per release`. After that I think this is good to be merged.

settings
ekohl: With `katello` I want `FOREMAN_VERSION=` in the generated file, regardless of whether it's empty or not. I didn't think of `plugins` where it's special.

ekohl: How about limiting it to just katello?
```suggestion
	if [[ $PROJECT == katello ]] ; then
```

And then move it into the existing katello block.

evgeni: Yeah, that seems to do it, thanks!

evgeni (commented):
acking non-evgeni parts of this PR ;)

ekohl (approved):
I'll squash the commits into what it's now.

PR #492, opened 2025-05-09
Automatic checkout of theforeman.org repository

### LINKED ISSUE #491: RFE: export_gpg_public - Automatic checkout of theforeman.org repository
Having experience with the `bump_packaging` script and knowing how nicely the automatic/magic checkout works, I would expect the `export_gpg_public` script to do the same. 

https://github.com/theforeman/theforeman-rel-eng/blob/dbf0003121ca38e0af01a25811ae7c9741358bb5/bump_packaging#L29-L31

It should be relatively easy to implement if we reach a consensus.

ogajduse:
Fixes #491


export_gpg_public
```diff
@@ -15,6 +15,13 @@ fi
 
 CHECKOUT="${GIT_DIR}/theforeman.org"
 
+if [[ ! -d "$CHECKOUT" ]] ; then
+	git clone --quiet --origin "$GIT_REMOTE" https://github.com/theforeman/theforeman.org "$CHECKOUT"
+fi
+
+git -C "$CHECKOUT" fetch "$GIT_REMOTE"
+git -C "$CHECKOUT" checkout -b "add-gpg-key-$VERSION" "$GIT_REMOTE/$TFM_WEBSITE_GH_PAGES_BRANCH"
```
ekohl, line 23: Is this variable defined? On my phone so annoying to check but I don't recall having this tfm var

ogajduse: Great catch! I forgot to commit my changes in the _settings_ file. Fixed in the latest version.

ekohl (approved):

ekohl:
Ignoring shellcheck. I think we need to start exporting those settings but for now I'm fine ignoring it.

PR #497, opened 2025-05-21
Inform the user that input is expected

lhellebr:


release_announcement
```diff
@@ -15,7 +15,7 @@ fi
 echo >> "$ANNOUNCEMENT"
 
 if [[ $PROJECT == foreman ]] ; then
-	# TODO: inform the user input is expected?
+	echo Enter the announcement text, end by CTRL+D:
```
ekohl, line 18: Technically if there already was something on stdin then it's not expected, but we can ignore that.

More important thing to consider: `read -p` prints on stderr so you can copy stdout (for example, using `wl-copy`). I'd like to maintain that option.

ekohl (changes requested):

release_announcement
lhellebr: `>&2`

release_announcement
```diff
@@ -15,7 +15,7 @@ fi
 echo >> "$ANNOUNCEMENT"
 
 if [[ $PROJECT == foreman ]] ; then
-	# TODO: inform the user input is expected?
+	echo Enter the announcement text, end by CTRL+D: >&2
```
ekohl, line 18: A very nitpicky thing: technically someone could pipe data in and `cat` will consume that. However, given the context that this script is used in I think it's fine to ignore that edge case now. At least the message it put on stderr so you can still copy the output with a pipe and that's more important.

ekohl (approved):

PR #498, opened 2025-05-21
Remove Twitter announcement

lhellebr:


ekohl (approved):

evgeni:
Should we instead share it on the Fediverse?

Issue #364, opened 2024-05-15
Manage the Redmine version TODO/DONE saved queries


ekohl:
We already have steps to create a new version:
https://github.com/theforeman/theforeman-rel-eng/blob/1b2ff8be0f43fb86f82b675bdd7fad78ba3c8ec2/procedures/foreman/release.md.erb#L20-L22

At some in the procedure we need to update the TODO/DONE queries (like https://projects.theforeman.org/projects/foreman/issues?query_id=171 and https://projects.theforeman.org/projects/foreman/issues?query_id=175) to reflect that. Probably good to wait until things have been released, or at the very least until the release notes have been created.

PR #553, opened 2026-02-10
Automate Redmine version creation with create_redmine_version script

ogajduse:
## Summary

Automate Redmine version creation by adding a new `create_redmine_version` script, eliminating the need for manual version creation through the web interface.

## Changes

- **New script**: `create_redmine_version` - Automates creation of Redmine versions following the same pattern as `close_redmine_version`
- **Updated procedures**: Both branch and release procedures now reference the automated script instead of manual steps

## Implementation Details

The `create_redmine_version` script:
- Uses `require_fullversion` to load version from settings hierarchy
- Strips any `-rc` suffix to get the base version (e.g., `3.18.0-rc1` → `3.18.0`)
- Leverages the existing `redmine` Python script with `version create`
- Automatically sets `--status open --sharing descendants`
- Prompts for Redmine API key if not already set in environment

## Benefits

- Reduces manual work and potential errors during branching and releases
- Consistent with existing automation patterns (`close_redmine_version`)
- No manual web UI interaction required
- Uses the same settings system as other release scripts

## Context

Addresses feedback from @ekohl in #517 to create a script similar to `close_redmine_version` for automating Redmine version creation.

Partially addresses #364 (automates version creation; saved query management remains to be implemented)

### DIFF
--- a/redmine
+++ b/redmine
@@ -57,7 +57,7 @@ def close_versions_by_pattern(args: argparse.Namespace) -> None:
     redmine = get_redmine()
     all_versions = list(redmine.version.filter(project_id=args.project_id))
 
-    # Filter for open versions (client-side since API filter is unreliable)
+    # Filter for open versions (client-side since API filter is not available)
     open_versions = [v for v in all_versions if v.status == 'open']
 
     if not open_versions:


procedures/foreman/release.md.erb
```diff
@@ -37,8 +40,7 @@
   - [ ] Push tags using <%= rel_eng_script('tag_push') %>
 - [ ] Run the Jenkins [Tarballs Release](https://ci.theforeman.org/job/tarballs-release/) using <%= rel_eng_script('release_tarballs') %> to create tarballs
 <% unless is_rc -%>
-- [ ] Change Redmine version <%= full_version %> state to Closed using <%= rel_eng_script('close_redmine_version') %>
```
ekohl: I had them close together for a reason: IMHO it should be almost atomic because we should never have no open version. The PR processor sets the fixed version in Redmine and relies on it.

Perhaps change the order to effectively:
* Tag projects
  * `./tag_project`
  * `./tag_push`
* Update Redmine
  * `./close_redmine_version`
  * `./create_redmine_version`
* `./release_tarballs`

pcreech: My vote _here_ is to drop the fallback, and make them both explicit arguments.  This implies a level of intentionality to the scripts that could be missed with environment variables.

pcreech:
If, at the moment, we can't stop using "VERSION" due to parameter checking, it may be better to keep using that format, due to expectations for scripts in the repo.  As much as I like the more intentional move with arguments, i don't want to break paradigms in here.

Now, if we have precedence in other areas of the repo, by all means, let's break settings file inheritance and switch to arguments only.

redmine
```diff
@@ -46,10 +48,51 @@ def set_status_version(args: argparse.Namespace) -> None:
...
 
 
+def close_versions_by_pattern(args: argparse.Namespace) -> None:
+    """Close all open versions matching a pattern.
+
+    Refuses to close if it would leave no open versions.
+    """
+    redmine = get_redmine()
+    all_versions = list(redmine.version.filter(project_id=args.project_id))
+    open_versions = [v for v in all_versions if v.status == 'open']
```
ekohl: I haven't tested this, but is it possible to filter server side?
```suggestion
    open_versions = list(redmine.version.filter(project_id=args.project_id, status='open'))
```

ekohl (commented):
Overall I think this looks good and wouldn't mind merging. Some minor detail inline.

PR #570, opened 2026-04-28
Move hyperlinked script out of the preformatted command

ogajduse:
This patch duplicates string_freeze_announcement and transifex_announcement to allow hyperlink as well as having the whole command preformatted.


ogajduse:
How it renders in Discourse:
image

ekohl (approved):

PR #582, opened 2026-05-25
Support candlepin in release procedure

ogajduse:
Add candlepin to the allowed project list for the release subcommand,
skip --foreman-version when not applicable, and fix the Copr repo name
in the candlepin release template.

### DIFF
--- a/procedure
+++ b/procedure
@@ -109,7 +109,7 @@ procedures = {
     end
 	end,
 	'release' => OptionParser.new do |opts|
-    opts.on('--project [PROJECT]', ['foreman', 'katello'],
+    opts.on('--project [PROJECT]', ['foreman', 'katello', 'candlepin'],
             'Project to generate release procedure for') do |project|
       context[:project] = project
     end
@@ -155,7 +155,7 @@ procedures = {
 def missing_required_option?(procedure, context)
   case context[:project]
   when 'candlepin'
-    context[:release].nil?
+    context[:release].nil? && context[:short_version].nil?
   when 'katello'
     context[:target_date].nil? || context[:owner].nil? || (procedure == 'release' && context[:foreman_version].nil?)
   when 'foreman'
--- a/procedure
+++ b/procedure
@@ -155,7 +155,12 @@ procedures = {
 def missing_required_option?(procedure, context)
   case context[:project]
   when 'candlepin'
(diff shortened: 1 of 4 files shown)


procedure
```diff
@@ -155,7 +155,7 @@ procedures = {
 def missing_required_option?(procedure, context)
   case context[:project]
   when 'candlepin'
-    context[:release].nil?
+    context[:release].nil? && context[:short_version].nil?
```
ekohl: The idea of the function is to verify user input but `short_version` is a derived variable so shouldn't it check `version`?
```suggestion
    context[:release].nil? && context[:version].nil?
```

But looking further, should it check the exact procedure?
```suggestion
    case procedure
    when 'branch'
      context[:release].nil?
    when 'release'
      context[:version].nil?
    end
```

procedures/candlepin/release.md.erb
```diff
@@ -1,9 +1,9 @@
 # Build Package
 
-- [ ] Acquire Candlepin SRPM for <%= release %>
-- [ ] Build into Copr: `copr build @theforeman/candlepin-staging-<%= release %> <SRPM>`
+- [ ] Acquire Candlepin SRPM for <%= short_version %>
```
ekohl: Shouldn't this be for the exact version?
```suggestion
- [ ] Acquire Candlepin SRPM for <%= version %>
```

procedures/candlepin/release.md.erb
```diff
@@ -1,9 +1,9 @@
...
+- [ ] Build into Copr: `copr build @theforeman/candlepin-<%= short_version %>-staging <SRPM>`
```
ekohl: Is it and idea to also suggest the likely filename?
```suggestion
- [ ] Build into Copr: `copr build @theforeman/candlepin-<%= short_version %>-staging candlepin-<%= full_version %>.src.rpm`
```

ekohl (commented):
Nice idea

procedures/candlepin/release.md.erb
ogajduse: changed to `candlepin-<%= full_version %>*.src.rpm` as we need to account for RPM `Release`.

ogajduse:
Thank you for the feedback. Addressed. Ready for review.

zjhuntin (approved):

PR #594, opened 2026-07-03
Add next-patch mode to create_redmine_version

ogajduse:
## Summary
- `create_redmine_version` stripped the `-rc` suffix from `FULLVERSION` to derive the Redmine version name. Correct for branching (creates `develop.0`), wrong for the "create new version" step after a release is tagged, since `FULLVERSION` still holds the version just released (e.g. `3.19.1`), not the upcoming patch (`3.19.2`).
- Add a `next` argument: bumps the patch component instead of using `FULLVERSION` as-is.
- Update `procedures/foreman/release.md.erb` so the generated release procedure invokes `./create_redmine_version next` for that step.
- Branching procedure (`branch.md.erb`) is unaffected — it still uses the default mode.

## Test plan
- [x] `bash -n create_redmine_version` — syntax OK
- [x] Verified patch-bump math: `3.19.1` → `3.19.2`, `3.19.0` → `3.19.1`
- [x] Verified default mode unchanged: `3.20.0-rc1` → `3.20.0`
- [x] Bad argument produces usage error
- [x] Rendered `./procedure release` for a mid-series patch (`3.19.2`) and first-GA (`3.20.0`) — both correctly show `./create_redmine_version next` with the right target version and EOL caveat only where expected
- [x] Rendered an RC (`3.20.0-rc1`) — confirmed "Update Redmine versions" section is skipped entirely (unchanged, pre-existing `unless is_rc` guard)

### DIFF
--- a/procedures/foreman/release.md.erb
+++ b/procedures/foreman/release.md.erb
@@ -38,7 +38,7 @@
 <% unless is_rc -%>
 - Update Redmine versions
   - [ ] Close old <%= short_version %>.x versions using <%= rel_eng_script('close_redmine_version') %>
-  - [ ] Create new version <%= next_patch_version %><%= ' (unless the series is EOL)' unless is_first_ga %> using <%= rel_eng_script('create_redmine_version') %>
+  - [ ] Create new version <%= next_patch_version %><%= ' (unless the series is EOL)' unless is_first_ga %> using <%= rel_eng_script('create_redmine_version') %>: `./create_redmine_version next`
 <% end -%>
 - [ ] Run the Jenkins [Tarballs Release](https://ci.theforeman.org/job/tarballs-release/) using <%= rel_eng_script('release_tarballs') %> to create tarballs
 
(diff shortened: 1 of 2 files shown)


procedures/foreman/release.md.erb
```diff
@@ -38,7 +38,7 @@
 <% unless is_rc -%>
 - Update Redmine versions
   - [ ] Close old <%= short_version %>.x versions using <%= rel_eng_script('close_redmine_version') %>
-  - [ ] Create new version <%= next_patch_version %><%= ' (unless the series is EOL)' unless is_first_ga %> using <%= rel_eng_script('create_redmine_version') %>
+  - [ ] Create new version <%= next_patch_version %><%= ' (unless the series is EOL)' unless is_first_ga %> using <%= rel_eng_script('create_redmine_version') %>: `./create_redmine_version next`
```
ekohl, line 41: `next_patch_version` is used in 2 places in the procedure so dropping it here won't simplify the code.

ekohl (approved):
I debated `create_redmine_version` vs `create_redmine_next_version` + `create_redmine_current_version`, but in the end I think this is still simple enough.

PR #595, opened 2026-07-22
procedures/candlepin: add container image release step

Odilhao:
## Summary
Adds a "Update Container Image" section to `procedures/candlepin/release.md.erb`, covering the step of bumping the Candlepin version pinned in [candlepin-oci-images](https://github.com/theforeman/candlepin-oci-images) branches after a Candlepin RPM release.

## Motivation
[theforeman-rel-eng-konflux](https://github.com/theforeman/theforeman-rel-eng-konflux)'s `discourse-oci-branching.md.j2` template covers the broader multi-repo Foreman OCI branching flow (foreman + pulp + candlepin together, for a *new* Foreman version). It doesn't cover the narrower, recurring case of a standalone Candlepin point release needing its existing `candlepin-oci-images` branch(es) bumped to pick up the new RPM — this fills that gap, scoped only to Candlepin.

## Verification
Rendered locally with the repo's own `procedure` script to confirm ERB syntax and output:
```
ruby ./procedure release --project candlepin --version 4.7.5 --target-date 2026-08-01 --owner testuser
```
Output includes the new "Update Container Image" section with `short_version`/`version` interpolated correctly.


procedures/candlepin/release.md.erb
```diff
@@ -8,3 +8,12 @@
...
 - [ ] Announce the release on [Discourse](https://community.theforeman.org/c/release-announcements/8) using <%= rel_eng_script('release_announcement') %>: `FULLVERSION=<%= version %> VERSION=<%= short_version %> PROJECT=candlepin ./release_announcement`
+
+# Update Container Image
+
+This bumps the Candlepin version pinned in [candlepin-oci-images](https://github.com/theforeman/candlepin-oci-images) branches that already track Candlepin <%= short_version %>. It does not create new branches — that happens automatically as part of a Foreman release's OCI branching, see [theforeman-rel-eng-konflux](https://github.com/theforeman/theforeman-rel-eng-konflux).
+
+- [ ] Identify branches pinned to Candlepin <%= short_version %>: `gh api repos/theforeman/candlepin-oci-images/branches --jq '.[].name' | while read -r b; do gh api "repos/theforeman/candlepin-oci-images/contents/images/candlepin/Containerfile?ref=$b" -q .content 2>/dev/null | base64 -d | grep -q "VERSION=<%= short_version %>$" && echo "$b"; done`
```
ekohl: I'd suggest to have a small helper script to capture that instead of inlining it here.

procedures/candlepin/release.md.erb
```diff
@@ -8,3 +8,12 @@
...
 - [ ] Kick off the [release pipeline](https://ci.theforeman.org/job/candlepin-<%= short_version %>-rpm-pipeline/) by calling <%= rel_eng_script('release_pipeline') %>: `VERSION=<%= short_version %> PROJECT=candlepin ./release_pipeline`
 - [ ] Announce the release on [Discourse](https://community.theforeman.org/c/release-announcements/8) using <%= rel_eng_script('release_announcement') %>: `FULLVERSION=<%= version %> VERSION=<%= short_version %> PROJECT=candlepin ./release_announcement`
+
+# Update Container Image
+
+This bumps the Candlepin version pinned in [candlepin-oci-images](https://github.com/theforeman/candlepin-oci-images) branches that already track Candlepin <%= short_version %>. It does not create new branches — that happens automatically as part of a Foreman release's OCI branching, see [theforeman-rel-eng-konflux](https://github.com/theforeman/theforeman-rel-eng-konflux).
+
+- [ ] Identify branches pinned to Candlepin <%= short_version %> using <%= rel_eng_script('candlepin_oci_branches') %>: `VERSION=<%= short_version %> ./candlepin_oci_branches`
```
ogajduse:
```suggestion
- [ ] Identify branches pinned to Candlepin <%= short_version %> using <%= rel_eng_script('candlepin_oci_branches') %>: `VERSION=<%= short_version %> PROJECT=candlepin ./candlepin_oci_branches`
```

procedures/candlepin/release.md.erb
```diff
@@ -8,3 +8,12 @@
...
+- [ ] On each affected branch, bump `VERSION_XYZ` in `images/candlepin/Containerfile` to `<%= version %>` and open a PR: `sed -i 's/^ARG VERSION_XYZ=.*/ARG VERSION_XYZ=<%= version %>/' images/candlepin/Containerfile`
```
ogajduse: Can we make the scripts in this project run operations in the candlepin-oci-images project the same way we do it with foreman-packaging? I was thinking about introducing new settings for that in `settings`.

```
CANDLEPIN_OCI_IMGS_DIR="$GIT_DIR/candlepin-oci-images"
CANDLEPIN_OCI_IMGS_GIT_REMOTE=origin
CANDLEPIN_OCI_IMGS_FORK_REMOTE="${PACKAGING_FORK_REMOTE:-${GITHUB_USER}}"
```

And the script(s) would use them.

Odilhao: Done — added `candlepin_oci_branches` (identifies affected branches via a local clone instead of `gh api`) and a new `bump_candlepin_oci_image` script that bumps `VERSION_XYZ`, commits, pushes to a fork, and opens a PR, mirroring `bump_packaging`'s flow for `foreman-packaging`.

Settings-wise I went with a slightly more generic naming than proposed, since this repo also has `foreman` and other projects, not just candlepin:

```
OCI_DIR="${OCI_DIR:-$GIT_DIR/${PROJECT}-oci-images}"
OCI_GIT_REMOTE="${OCI_GIT_REMOTE:-origin}"
OCI_FORK_REMOTE="${OCI_FORK_REMOTE:-${GITHUB_USER}}"
```

`OCI_DIR` is derived from `$PROJECT` so it isn't candlepin-specific, and all three follow the existing `${VAR:-default}` override pattern (like `GIT_REMOTE`) instead of being hardcoded.

bump_candlepin_oci_image
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash -e
+
+. settings
+
+require_fullversion
```
ogajduse, line 5: `VERSION` selects which branches get bumped (via `candlepin_oci_branches`) and `FULLVERSION` is what gets written into them, but nothing checks the two agree. `VERSION=4.8 FULLVERSION=4.7.5` would find the 4.8-pinned branches and write `ARG VERSION=4.8` + `ARG VERSION_XYZ=4.7.5`, producing an RPM URL that 404s at build time rather than a visible error here.

A guard next to `require_fullversion` catches it, in the spirit of the existing `require_*` helpers:

```suggestion
require_fullversion

if [[ $FULLVERSION != $VERSION.* ]] ; then
  echo "FULLVERSION ($FULLVERSION) is not a release of VERSION ($VERSION)" >&2
  exit 1
fi
```

bump_candlepin_oci_image
```diff
@@ -0,0 +1,51 @@
...
+
+REPO="${PROJECT}-oci-images"
+CONTAINERFILE=images/candlepin/Containerfile
```
ogajduse: This is half-generic in a way that will bite eventually: `REPO` and the `OCI_*` settings all derive from `$PROJECT`, but this path, the commit/PR titles, and the script names are hardcoded to candlepin. `PROJECT=foreman ./bump_candlepin_oci_image` would clone `foreman-oci-images` and then go looking for `images/candlepin/Containerfile`.

Either commit to generic (`CONTAINERFILE="images/${PROJECT}/Containerfile"`, `"Bump ${PROJECT^} to $FULLVERSION"`, rename to `oci_branches` / `bump_oci_image`) or drop `$PROJECT` from `REPO` and the `OCI_*` defaults and hardcode candlepin throughout. Right now it's neither, which is the option that silently misbehaves.

bump_candlepin_oci_image
```diff
@@ -0,0 +1,51 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+BRANCHES=$("$SCRIPT_DIR/candlepin_oci_branches")
```
ogajduse: `SCRIPT_DIR`/`BASH_SOURCE` doesn't appear anywhere else in this repo &mdash; every script that calls a sibling uses a bare relative path (`branch_push:5`, `tag_project:7`, `bump_rpm_packaging:3`, `sign_stage_rpms:11`, ...). It also doesn't buy anything here: `. settings` on line 3 is already a relative source, so the script only works from the repo root either way.

```suggestion
BRANCHES=$(./candlepin_oci_branches)
```

bump_candlepin_oci_image
```diff
@@ -0,0 +1,51 @@
...
+if [[ -z "$BRANCHES" ]] ; then
+  exit 0
+fi
+
+cd "$OCI_DIR"
+
+for branch in $BRANCHES; do
+  FEATURE_BRANCH="bump-candlepin-${FULLVERSION}-${branch//\//-}"
+
+  git checkout -B "$FEATURE_BRANCH" "$OCI_GIT_REMOTE/$branch"
```
ogajduse, line 44: `-B` force-resets an existing feature branch so a re-run gets past this line, but then dies at line 25 with "nothing to commit" (the sed is already applied) or at `gh pr create` because a PR already exists for that head.

Combined with the loop, that means re-running to pick up a branch that was missed the first time isn't possible without manual cleanup. Worth either skipping branches that already have an open bump PR, or checking `git ls-remote` for the feature branch up front.

bump_candlepin_oci_image
```diff
@@ -0,0 +1,51 @@
...
+  sed -i "s/^ARG VERSION_XYZ=.*/ARG VERSION_XYZ=${FULLVERSION}/" "$CONTAINERFILE"
```
ogajduse, line 53: Two things on this line.

**The sed can silently do nothing.** `candlepin_oci_branches` only greps for `ARG VERSION=`; it never confirms `ARG VERSION_XYZ=` exists on the matched branch. If the sed matches nothing, `git add` stages nothing, `git commit` fails, and `-e` kills the script **mid-loop** &mdash; branch 1 has a PR, branches 2..N are silently untouched, and there's no summary of what happened. `bump_packaging` has the same shape but only ever handles one branch, so this is a new failure mode. `ensure_clean_git_checkouts:18` already uses the check you'd want:

```suggestion
  sed -i "s/^ARG VERSION_XYZ=.*/ARG VERSION_XYZ=${FULLVERSION}/" "$CONTAINERFILE"

  if git diff --quiet -- "$CONTAINERFILE" ; then
    echo "$branch: no ARG VERSION_XYZ change in $CONTAINERFILE" >&2
    exit 1
  fi
```

**The `Makefile` needs bumping too** &mdash; see the summary comment. `PROJECT_XYZ_TAG` is what drives the published tags, and it has already drifted to `4.8.2` on master while the Containerfile says `4.8.4`.

bump_candlepin_oci_image
```diff
@@ -0,0 +1,51 @@
...
+  git add "$CONTAINERFILE"
+  git commit --message "Bump Candlepin to $FULLVERSION"
+
+  if [[ $PACKAGING_PR == true ]] ; then
+    # Only fork if it doesn't exist yet
+    if ! gh repo view "${GITHUB_USER}/${REPO}" --json url --jq .url &>/dev/null ; then
+      GH_PROMPT_DISABLED=1 gh repo fork "theforeman/${REPO}" --remote-name "$OCI_FORK_REMOTE"
+    fi
```
ogajduse: This fork/remote setup runs on every iteration of the loop. `bump_packaging` does it once because it only ever handles a single branch &mdash; worth hoisting it above the `for`.

Separately, `theforeman/` is hardcoded here and on lines 30 and 43, while `settings` has `GITHUB_NAMESPACE` and a `github_url()` helper that `candlepin_oci_branches:9` in this same PR already uses. `bump_packaging` hardcodes it too, so this is a copied wart rather than a new one, but it's inconsistent within the PR itself.

ogajduse (commented):
Reviewed at c1a6989. I compared the new scripts against the existing ones (`bump_packaging`, `ensure_clean_git_checkouts`, `tag_project`, ...) and checked the behaviour against `candlepin-oci-images` and `theforeman-rel-eng-konflux`.

The shape mirrors `bump_packaging` closely and the ERB renders correctly. Both new scripts also pass Differential ShellCheck clean, which `bump_packaging` itself doesn't (SC2086 on line 42). One blocking issue, plus hardening and consistency points inline.

## The version pin lives in two files; only one is bumped

`candlepin-oci-images` pins the Candlepin version in **both** places:

- `images/candlepin/Containerfile` &mdash; `ARG VERSION` / `ARG VERSION_XYZ`
- `Makefile` &mdash; `PROJECT_XY_TAG` / `PROJECT_XYZ_TAG`, which feed `IMAGE_TAGS` and the `--build-arg`s in `make build`

The konflux runbook treats them as a pair: `docs/oci-branching-runbook.md` Phase 3.5 patches the Containerfile, Phase 3.6 the Makefile. `bump_candlepin_oci_image` only patches the Containerfile, and the drift is already live on `master`:

| file | value on master |
| --- | --- |
| `Containerfile` &rarr; `ARG VERSION_XYZ` | `4.8.4` |
| `Makefile` &rarr; `PROJECT_XYZ_TAG` | `4.8.2` |

Every bump up to and including 4.8.2 changed both files in a single commit. The last two &mdash; 4.8.3, and theforeman/candlepin-oci-images@b4e02d6 "Bump Candlepin to 4.8.4" from branch `bump-candlepin-4.8.4-master` (this script's `FEATURE_BRANCH` pattern, per the test run posted above) &mdash; touched the Containerfile only. So merging as-is bakes the drift in rather than fixing it.

`theforeman-rel-eng-konflux`'s `hack/branch-release/lib/containerfile.py` already has a `patch_arg()` that does idempotent ARG patching and errors when the ARG is missing. Worth at least matching its semantics.

## Which registry actually carries the released tags?

`.tekton/candlepin-3-19-push.yaml` sets `output-image: quay.io/foreman/stage/candlepin:{{revision}}` &mdash; a stage registry keyed by git revision. The semantic tags on `quay.io/foreman/candlepin` come from the Konflux ReleasePlan, not from merging to `candlepin-oci-images`.

`candlepin_oci_tag` inspects `quay.io/foreman/candlepin` and compares `vcs-ref` against the branch HEAD, which assumes promotion has already happened and preserves the label. Worth confirming against the ReleasePlan &mdash; if promotion lags the merge, line 42's `NOT YET` becomes the normal outcome rather than the exception, and it stops being distinguishable from a genuinely failed build.

## Procedure placement

"Update Container Image" lands after "Announce the release on Discourse", so the release is announced before the container carrying it exists. That may well be deliberate &mdash; the RPMs *are* the release &mdash; but it reads like an artifact of appending to the end of the file, so worth making an explicit call.

candlepin_oci_branches
```diff
@@ -0,0 +1,34 @@
...
+CONTAINERFILE=images/candlepin/Containerfile
+
+if [[ ! -d "$OCI_DIR" ]] ; then
+  git clone --quiet --origin "$OCI_GIT_REMOTE" "$(github_url "$REPO")" "$OCI_DIR"
+fi
+
+(
+  cd "$OCI_DIR"
+  git fetch --quiet "$OCI_GIT_REMOTE"
+)
```
ogajduse: Two small things: the subshell `cd` is unnecessary given lines 23 and 29 already use `git -C`, and the fetch doesn't prune.

Without `--prune`, a stale remote-tracking ref for an upstream branch that has since been deleted still gets emitted by line 29, `checkout -B` succeeds against it, and then `gh pr create --base` fails on a base that no longer exists.

```suggestion
git -C "$OCI_DIR" fetch --prune --quiet "$OCI_GIT_REMOTE"
```

candlepin_oci_branches
```diff
@@ -0,0 +1,34 @@
...
+
+found=false
+while read -r branch; do
+  case "$branch" in
+    HEAD|dependabot/*|konflux/*|use-*) continue ;;
+  esac
+
+  content=$(git -C "$OCI_DIR" show "$OCI_GIT_REMOTE/$branch:$CONTAINERFILE" 2>/dev/null) || continue
+
+  if echo "$content" | grep -q "ARG VERSION=${VERSION}\$"; then
```
ogajduse: `$VERSION` goes into a BRE unescaped, so `.` matches any character (`4.8` would also match `428`), and the pattern isn't anchored at the start, so a commented-out `# ARG VERSION=4.8` would match too.

Neither is likely against today's Containerfiles, but the entire branch-selection logic hangs on this one line. A fixed-string exact match is both stricter and simpler:

```suggestion
  if echo "$content" | grep -qFx "ARG VERSION=${VERSION}"; then
```

candlepin_oci_branches
```diff
@@ -0,0 +1,34 @@
...
+    echo "$branch"
+    found=true
+  fi
+done < <(git -C "$OCI_DIR" branch --remotes --list "$OCI_GIT_REMOTE/*" --format='%(refname:lstrip=3)')
+
+if [[ $found == false ]]; then
+  echo "No $OCI_GIT_REMOTE branch of $REPO is pinned to VERSION=${VERSION} yet." >&2
+  echo "It's likely not adopted anywhere yet and will land on master first." >&2
+fi
```
ogajduse, line 32: The second line is speculation rather than an actionable message, which doesn't quite match how the rest of the repo reports problems.

More to the point: finding nothing exits 0, and `bump_candlepin_oci_image:13` then exits 0 as well &mdash; so a typo'd `VERSION` is indistinguishable from success for anyone running the bump step on its own.

candlepin_oci_tag
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash -e
+
+. settings
+
+REPO="${PROJECT}-oci-images"
+IMAGE="quay.io/foreman/${PROJECT}"
```
ogajduse, line 11: See the summary comment: `.tekton/candlepin-3-19-push.yaml` publishes to `quay.io/foreman/stage/candlepin:{{revision}}`, so this is inspecting the post-promotion location. Worth confirming the Konflux ReleasePlan actually promotes to `quay.io/foreman/candlepin` and preserves the `vcs-ref` label &mdash; otherwise line 42's `NOT YET` is the expected output rather than the exception.

Also `quay.io/foreman` is hardcoded while the image name interpolates `$PROJECT`, which is the same half-generic issue as in `bump_candlepin_oci_image`.

candlepin_oci_tag
```diff
@@ -0,0 +1,44 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+BRANCHES=$("$SCRIPT_DIR/candlepin_oci_branches")
+
+if [[ -z "$BRANCHES" ]] ; then
+  exit 0
+fi
```
ogajduse: This preamble (`SCRIPT_DIR`, call `candlepin_oci_branches`, bail if empty) is now duplicated verbatim from `bump_candlepin_oci_image:9-15`, so the `SCRIPT_DIR` point applies here too.

Worth noting that `candlepin_oci_branches` clones and fetches as a side effect, which is a little surprising for a script whose job is read-only verification of what's already published.

candlepin_oci_tag
```diff
@@ -0,0 +1,44 @@
...
+    TAG=nightly
+  else
+    TAG="$VERSION"
+  fi
+
+  EXPECTED_COMMIT=$(git -C "$OCI_DIR" rev-parse "$OCI_GIT_REMOTE/$branch")
+
+  echo "== $branch (tag: $TAG) =="
+
+  if ! INSPECT=$(skopeo inspect "docker://${IMAGE}:${TAG}" 2>/dev/null) ; then
```
ogajduse, line 26: `skopeo` isn't used anywhere else in this repo, and `jq` currently only appears as `gh --jq` (gh's built-in, not the binary) &mdash; so this adds two new hard dependencies.

Worth listing both in the README alongside the existing gh / ruby / gpg / gopass requirements, and failing early with a clear message when `skopeo` is missing rather than at the first inspect.

bump_candlepin_oci_image
ogajduse: This is optional. I think our settings for candlepin releases might need some more thought about how we handle `VERSION` and `FULLVERSION`.

candlepin_oci_tag
ogajduse: I think cloning and fetching the repository are fine in this case.

bump_candlepin_oci_image
Odilhao: Left this out for now — you flagged it as optional yourself and it ties into the broader VERSION/FULLVERSION settings question you raised, which is worth its own discussion rather than a quick guard here.

bump_candlepin_oci_image
Odilhao: Fixed: `CONTAINERFILE` now derives from `$PROJECT` (`images/${PROJECT}/Containerfile`) and commit/PR text uses `"Bump ${PROJECT^} to $FULLVERSION"` instead of a hardcoded "Candlepin". Kept the script names as-is since they're invoked with `PROJECT=candlepin` explicitly from the procedure, same as the other candlepin-specific steps in that file.

bump_candlepin_oci_image
Odilhao: Fixed — switched to `./candlepin_oci_branches` in both scripts, dropped `SCRIPT_DIR`.

bump_candlepin_oci_image
Odilhao: Fixed: before doing any work for a branch, checks `gh pr list --head $GITHUB_USER:$FEATURE_BRANCH --state open` and skips with a message if one's already open, so a re-run to pick up a branch that failed earlier doesn't touch branches that already have a PR out.

bump_candlepin_oci_image
Odilhao: Good catch on both counts.\n\nFor the sed: added a `git diff --quiet` check on both files after the sed. Went slightly further than the suggested snippet though — instead of `exit 1` immediately (which still leaves branches 2..N untouched with no summary, per your own point), it `continue`s past the failed branch, collects failures in an array, and reports them all at the end with a non-zero exit. So one branch failing doesn't stop the others from getting bumped.\n\nFor the Makefile: now bumping `PROJECT_XY_TAG`/`PROJECT_XYZ_TAG` alongside the Containerfile in the same commit. Verified against a scratch clone that the sed handles both the drifted form (`PROJECT_XYZ_TAG=\.2`, currently on master) and the literal form (`PROJECT_XYZ_TAG=4.7.5`, on foreman-3.19) correctly, and that the no-op case (value already matches) is correctly caught by the diff-guard.

bump_candlepin_oci_image
Odilhao: Fixed: hoisted above the loop, and switched to `${GITHUB_NAMESPACE}`/`github_url()` instead of hardcoded `theforeman/`, matching `candlepin_oci_branches`.

candlepin_oci_branches
Odilhao: Fixed both: dropped the subshell in favor of `git -C`, and added `--prune` to the fetch.

candlepin_oci_branches
Odilhao: Fixed, switched to `grep -qFx` for an exact fixed-string whole-line match.

candlepin_oci_branches
Odilhao: Agreed on both. Reworded to be actionable, and finding nothing now exits 1 instead of 0 — verified that `set -e` propagates that failure correctly through `BRANCHES=$(./candlepin_oci_branches)` in both `bump_candlepin_oci_image` and `candlepin_oci_tag` (a plain command-substitution assignment isn't exempt from `-e`, unlike a condition in `if`/`while`), so a typo'd VERSION now fails loudly in both.

candlepin_oci_tag
Odilhao: Checked this concretely: `.tekton/candlepin-3-19-push.yaml`'s `output-image` does go to `quay.io/foreman/stage/candlepin:{{revision}}` as you say. But I compared that staging tag against the promoted one directly — `quay.io/foreman/stage/candlepin:4bffa3ea5545e4f291675fb92a535865534c50e8` and `quay.io/foreman/candlepin:nightly` currently resolve to the exact same digest (`sha256:14c9fb23d92...`) with the same `vcs-ref` label. So the ReleasePlan promotes by digest and preserves the label intact — `candlepin_oci_tag` inspecting `quay.io/foreman/candlepin` directly is checking the right, already-promoted location, and `vcs-ref` does survive promotion. Left the code as-is given that evidence, but happy to add a comment noting this if useful.\n\nOn hardcoded `quay.io/foreman`: this one's intentional, not the same issue as `bump_candlepin_oci_image`. `foreman` is the actual, fixed Quay org that every project (candlepin, foreman, pulp) publishes under regardless of `$PROJECT` — it's not a per-project value like the OCI git repo name, so there's nothing to generalize there. `$PROJECT` is already interpolated for the image name itself (`quay.io/foreman/${PROJECT}`), which is the part that actually varies.

candlepin_oci_tag
Odilhao: Fixed the SCRIPT_DIR duplication, same as the other script.\n\nOn cloning/fetching being a side effect of a read-only verification script: agreed with your follow-up reply that it's fine here — it's the same clone `bump_candlepin_oci_image` already made/uses, so there's no meaningfully separate side effect being introduced.

candlepin_oci_tag
Odilhao: Fixed: `candlepin_oci_tag` now checks `command -v skopeo` up front and exits with a clear message if missing, instead of failing at the first inspect. Documented both `skopeo` and `jq` in the README's Setup section alongside the GitHub CLI requirement.

ogajduse (commented):
Re-reviewed at c5c6d3b2. Confirmed fixed in the code, not just claimed: `SCRIPT_DIR` gone from both scripts, fork/remote setup hoisted out of the loop and collapsed to `github_url "$REPO"` (cleaner than `bump_packaging`'s inline ssh/https branch), `GITHUB_NAMESPACE` throughout, PR-already-open check, failure collection with non-zero exit, `--prune`, `grep -qFx`, reworded stderr with `exit 1`, both now-dead `-z "$BRANCHES"` blocks removed, `skopeo` guard, README updated. The `set -e` reasoning in your reply is right too &mdash; `VAR=$(cmd)` is a simple command whose status is `cmd`'s, so it propagates.

Two things on the Makefile fix itself, though &mdash; one of them defeats its own purpose. Details inline.

Also: the digest comparison you did answers the stage-vs-promoted registry question properly, and you're right that `quay.io/foreman` is a fixed org rather than a per-project value, so that one's settled. Not re-raising either.

bump_candlepin_oci_image
```diff
@@ -0,0 +1,68 @@
...
+    continue
+  fi
+
+  git checkout -B "$FEATURE_BRANCH" "$OCI_GIT_REMOTE/$branch"
+
+  sed -i "s/^ARG VERSION_XYZ=.*/ARG VERSION_XYZ=${FULLVERSION}/" "$CONTAINERFILE"
+  sed -i "s/^PROJECT_XY_TAG=.*/PROJECT_XY_TAG=${VERSION}/" Makefile
+  sed -i "s/^PROJECT_XYZ_TAG=.*/PROJECT_XYZ_TAG=${FULLVERSION}/" Makefile
+
+  if git diff --quiet -- "$CONTAINERFILE" || git diff --quiet -- Makefile ; then
```
ogajduse: This needs to be `&&`, not `||`.

`git diff --quiet` exits 0 when there are **no** changes, so as written this reads *"Containerfile unchanged **or** Makefile unchanged &rarr; mark failed and skip"*. It should only bail when *neither* file needed bumping.

The practical effect is that the script can't repair a Containerfile/Makefile mismatch &mdash; which is exactly the state the Makefile fix was added to address. Run it against `master` today (`FULLVERSION=4.8.4 VERSION=4.8`) to bring the Makefile up from 4.8.2:

- `Containerfile` is already `4.8.4` &rarr; unchanged
- `Makefile` `PROJECT_XYZ_TAG` goes `${PROJECT_XY_TAG}.2` &rarr; `4.8.4` &rarr; changed
- `true || false` &rarr; branch lands in `FAILED`, gets skipped, script exits 1

The clean forward path (4.8.5 from a consistent state) works fine, so this won't surface in a happy-path test &mdash; only when the two files are out of sync.

```suggestion
  if git diff --quiet -- "$CONTAINERFILE" && git diff --quiet -- Makefile ; then
```

Worth flagging that `&&` is the minimal fix rather than the complete one: `git diff` can't distinguish *"already at the target value"* from *"the pattern wasn't found at all"*, and only the second is actually an error. With `&&`, a branch whose Containerfile is missing `ARG VERSION_XYZ` entirely would still commit a Makefile-only bump and look like it succeeded. Checking that the ARG/var is present before sed'ing separates the two cases &mdash; which is what `containerfile.py:patch_arg()` does (errors on missing, no-ops silently on already-correct).

bump_candlepin_oci_image
```diff
@@ -0,0 +1,68 @@
...
+       | grep -q true ; then
+    echo "$branch: PR already open for $FEATURE_BRANCH, skipping" >&2
+    continue
+  fi
+
+  git checkout -B "$FEATURE_BRANCH" "$OCI_GIT_REMOTE/$branch"
+
+  sed -i "s/^ARG VERSION_XYZ=.*/ARG VERSION_XYZ=${FULLVERSION}/" "$CONTAINERFILE"
+  sed -i "s/^PROJECT_XY_TAG=.*/PROJECT_XY_TAG=${VERSION}/" Makefile
+  sed -i "s/^PROJECT_XYZ_TAG=.*/PROJECT_XYZ_TAG=${FULLVERSION}/" Makefile
```
ogajduse, line 55: These two seds are unconditional, but the var names aren't universal across the OCI repos now that `CONTAINERFILE` is generic (`images/${PROJECT}/Containerfile`):

- `candlepin-oci-images` &mdash; has `PROJECT_XY_TAG` / `PROJECT_XYZ_TAG` &#10003;
- `pulp-oci-images` &mdash; has both &#10003;
- `foreman-oci-images` &mdash; has **neither**; its Makefile uses `KATELLO_VERSION` and `FOREMAN_XY_TAG` / `FOREMAN_XYZ_TAG`

So `PROJECT=foreman` would match nothing here, and combined with the `||` on line 43 every branch ends up in `FAILED`. The Containerfile path got genericized but the Makefile handling still assumes the candlepin/pulp layout, so the half-generic issue moved rather than went away.

Either gate the Makefile step on the vars actually being present, or drop the `$PROJECT` genericization and keep these scripts candlepin-scoped &mdash; both are fine, but the current middle ground is the one that fails quietly.

bump_candlepin_oci_image
```diff
@@ -0,0 +1,68 @@
...
+
+  if ! git remote get-url "$OCI_FORK_REMOTE" &>/dev/null ; then
+    git remote add "$OCI_FORK_REMOTE" "$(github_url "$REPO")"
+  fi
+fi
+
+FAILED=()
+
+for branch in $BRANCHES; do
+  FEATURE_BRANCH="bump-candlepin-${FULLVERSION}-${branch//\//-}"
```
ogajduse, line 35: Cosmetic, same inconsistency in miniature: this still hardcodes `bump-candlepin-` while the commit message and PR title moved to `${PROJECT^}`.

```suggestion
  FEATURE_BRANCH="bump-${PROJECT}-${FULLVERSION}-${branch//\//-}"
```

(Only worth applying if you keep the genericization &mdash; if you go the other way on the comment above, leave this as-is.)

bump_candlepin_oci_image
Odilhao: Good catch, and a bad one on my part — that inverted the guard's entire purpose. Fixed by collapsing to a single `git diff --quiet -- "$CONTAINERFILE" Makefile` (quiet only when neither path changed), which is simpler than either `&&` or `||` chains.

Went further on the deeper point too: added a `require_line()` presence check for `ARG VERSION_XYZ=`, `PROJECT_XY_TAG=`, and `PROJECT_XYZ_TAG=` before the sed runs, so a genuinely missing pattern (real error → FAILED) is now distinguished from "already at the target value" (reported and skipped, not an error). Verified all three cases against a scratch clone: today's actual master drift now commits correctly, an already-fully-applied branch is treated as a no-op rather than a failure, and a Containerfile missing `ARG VERSION_XYZ` entirely is caught by `require_line`.

bump_candlepin_oci_image
Odilhao: You're right, and this is exactly the kind of thing that fails quietly. Went with dropping the genericization rather than gating on presence: reverted `CONTAINERFILE` to a literal `images/candlepin/Containerfile` and the commit/PR text back to a literal "Candlepin". `REPO`/`OCI_*` still derive from `$PROJECT` since that naming pattern (`${PROJECT}-oci-images`) genuinely does hold across all three OCI repos, but the actual file layout these two scripts patch is candlepin-specific, so pretending otherwise was the bug.

bump_candlepin_oci_image
Odilhao: Went the other way per your parenthetical — reverted the genericization instead of extending it, so `FEATURE_BRANCH` staying hardcoded to `bump-candlepin-` is now consistent with `CONTAINERFILE` and the commit/PR text, not inconsistent with them.

bump_candlepin_oci_image
```diff
@@ -0,0 +1,81 @@
...
+CONTAINERFILE=images/candlepin/Containerfile
+
+BRANCHES=$(./candlepin_oci_branches)
+
+require_line() {
+  if ! grep -q "^${2}" "$1" ; then
+    echo "$3: $2 not found in $1" >&2
+    return 1
+  fi
+}
```
ogajduse, line 17: `require_fullversion` and `require_gpg_key` in `settings` take no arguments and `exit` on failure. This one takes three and `return`s, deliberately, so the loop can skip the branch and carry on &mdash; `check_line` signals that rather than implying the `require_*` contract. Also moves the definition above its first use.

```suggestion
check_line() {
  if ! grep -q "^${2}" "$1" ; then
    echo "$3: $2 not found in $1" >&2
    return 1
  fi
}

BRANCHES=$(./candlepin_oci_branches)
```

bump_candlepin_oci_image
```diff
@@ -0,0 +1,81 @@
...
+       | grep -q true ; then
+    echo "$branch: PR already open for $FEATURE_BRANCH, skipping" >&2
+    continue
+  fi
+
+  git checkout -B "$FEATURE_BRANCH" "$OCI_GIT_REMOTE/$branch"
+
+  if ! require_line "$CONTAINERFILE" "ARG VERSION_XYZ=" "$branch" \
+     || ! require_line Makefile "PROJECT_XY_TAG=" "$branch" \
+     || ! require_line Makefile "PROJECT_XYZ_TAG=" "$branch" ; then
```
ogajduse, line 48: `||` short-circuits, so a branch missing two of the three patterns only reports the first &mdash; you fix it, re-run, and find the next. Accumulating reports them all in one pass.

```suggestion
  missing=0
  check_line "$CONTAINERFILE" "ARG VERSION_XYZ=" "$branch" || missing=1
  check_line Makefile "PROJECT_XY_TAG=" "$branch"          || missing=1
  check_line Makefile "PROJECT_XYZ_TAG=" "$branch"         || missing=1

  if (( missing )) ; then
```

ogajduse (approved):
Approving &mdash; this is in good shape.

One optional nit inline, in two parts: `||` short-circuits so only the first missing pattern gets reported, and `require_*` in `settings` means "exit on failure" while this one returns so the loop can continue, hence the rename. Diagnostics only &mdash; I ran both versions end-to-end against a local fixture with `PACKAGING_PR=false`, and every case produced identical commits, pushes, skips and exit codes, with one extra stderr line when a branch is missing more than one pattern.

Take the suggestions if you like them, or merge as-is and I'll open an issue to sort it out separately. Your call &mdash; don't block on it.

procedures/candlepin/release.md.erb
```diff
@@ -8,3 +8,12 @@
...
 - [ ] Announce the release on [Discourse](https://community.theforeman.org/c/release-announcements/8) using <%= rel_eng_script('release_announcement') %>: `FULLVERSION=<%= version %> VERSION=<%= short_version %> PROJECT=candlepin ./release_announcement`
+
+# Update Container Image
+
+This bumps the Candlepin version pinned in [candlepin-oci-images](https://github.com/theforeman/candlepin-oci-images) branches that already track Candlepin <%= short_version %>. It does not create new branches — that happens automatically as part of a Foreman release's OCI branching, see [theforeman-rel-eng-konflux](https://github.com/theforeman/theforeman-rel-eng-konflux).
+
+- [ ] Identify branches pinned to Candlepin <%= short_version %> using <%= rel_eng_script('candlepin_oci_branches') %>: `VERSION=<%= short_version %> PROJECT=candlepin ./candlepin_oci_branches`
+- [ ] Bump `VERSION_XYZ` and open a PR on each affected branch using <%= rel_eng_script('bump_candlepin_oci_image') %>: `FULLVERSION=<%= version %> VERSION=<%= short_version %> PROJECT=candlepin ./bump_candlepin_oci_image`
```
ekohl, line 17: As noted in https://github.com/theforeman/theforeman-rel-eng/issues/597#issuecomment-5165356147 I think passing in `FULLVERSION` is not how the repo is supposed to be used today. In https://github.com/theforeman/theforeman-rel-eng/issues/598#issuecomment-5165226723 I also suggested that my preference is to adopt the same model as Foreman: the release process starts with a PR that bumps the `FULLVERSION`. Then from there you only pass in `VERSION`. The `bump_candlepin_oci_image` will automatically inherit the variable from the settings.

Alternatively, you never pass in `VERSION` and only `FULLVERSION`, but that's a bigger refactor.

ekohl (commented):
Looking ahead, how much of this will also apply to Pulpcore?

In other words, can the scripts be written as `bump_oci` and `tag_oci` with variables set in `releases/candlepin/settings` and `releases/pulpcore/settings` to make them work generically? You already use `PROJECT` in some places so generalizing the file naming is IMHO a nice thing.

PR #601, opened 2026-08-04
Fix module creation on build_stage_repository

Odilhao:
EL10 don't have modules anymore, so we don't need to build modules there

### DIFF
--- a/build_stage_repository
+++ b/build_stage_repository
@@ -348,7 +348,7 @@ def update_repositories(collection, version, dist, rpm_dir, srpm_dir, foreman_ve
     create_repository(rpm_dir)
     create_repository(srpm_dir)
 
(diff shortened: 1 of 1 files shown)


ogajduse (approved):

build_stage_repository
```diff
@@ -348,7 +348,7 @@ def update_repositories(collection, version, dist, rpm_dir, srpm_dir, foreman_ve
     create_repository(rpm_dir)
     create_repository(srpm_dir)
 
-    if collection in ['foreman'] or (collection in ['katello'] and dist == 'el8'):
+    if dist != 'el10' and (collection in ['foreman'] or (collection in ['katello'] and dist == 'el8')):
```
ekohl, line 351: Late, but perhaps simplify it down to just what we need these days?
```suggestion
    if collection == 'foreman' and dist == 'el9':
```

PR #603, opened 2026-08-10
Exclude foreman-installer from the el10 stage repository

ogajduse:
## Summary

Foreman 5.0 ships everything for el10 except `foreman-installer` and `foreman-installer-katello`, which aren't ready for that dist yet. The el10 comps file isn't a good place to enforce this since it's also consumed for `dnf group install` package selection on real systems, so this filters the two packages out directly in `build_stage_repository`.

- Adds an `EXCLUDED_PACKAGES` map keyed by `(collection, version, dist)`, currently covering `foreman`/`nightly`/`el10` and `foreman`/`5.0`/`el10`. An entry can be dropped once the excluded package(s) are ready for that dist, without affecting other versions.
- Binary RPMs: excluded packages are dropped from the comps-derived include set before download, so they're skipped both on fresh Copr downloads and cleaned up via the existing filter pass if already present from a prod-repo sync.
- Source RPMs: previously not filtered by comps at all, so a new `exclude_packages()` helper removes the excluded packages' SRPMs after download, covering both freshly-downloaded and prod-synced copies.

## Test plan

- [x] `python3 -m py_compile build_stage_repository`
- [x] Run `generate_stage_repository` (or `build_stage_repository foreman 5.0 el10`) for a Foreman 5.0 el10 stage and confirm `foreman-installer`/`foreman-installer-katello` are absent from ~both the binary and~ source repos, while other el10 packages are unaffected
- [x] Confirm el9 (and other dists/collections) behavior is unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/build_stage_repository
+++ b/build_stage_repository
@@ -21,8 +21,17 @@ if os.path.exists('/usr/bin/dnf4'):
 else:
   DNF = 'dnf'
 
+# Packages to keep out of a (collection, version, dist) stage repository even
+# though they are present in comps and get built in Copr. Remove an entry
+# once the excluded package(s) are ready to ship for that dist.
+EXCLUDED_PACKAGES = {
+    ('foreman', 'nightly', 'el10'): {'foreman-installer', 'foreman-installer-katello'},
+    ('foreman', '5.0', 'el10'): {'foreman-installer', 'foreman-installer-katello'},
+}
 
-def filter_packages(target_dir, packages):
+
+def filter_packages(target_dir, packages, excluded_packages=None):
+    excluded_packages = excluded_packages or set()
     files = glob.glob(f"{target_dir}/*.rpm")
 
     for file in files:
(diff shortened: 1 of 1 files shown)


zjhuntin (approved):
I think this makes a lot of sense with what we have, and works for what we'll need overtime. LGTM, thanks Ondrej!

Odilhao (approved):

ekohl (changes requested):
> The el10 comps file isn't a good place to enforce this since it's also consumed for `dnf group install` package selection on real systems, so this filters the two packages out directly in `build_stage_repository`.

IIRC our use of comps is different than standard RPM. Not listing it in the comps file is exactly what you should do. So really, https://github.com/theforeman/foreman-packaging/blob/rpm/develop/comps/comps-foreman-el10.xml shouldn't list `foreman-installer`. That's how we've historically filtered out packages.

PR #607, opened 2026-08-19
Pin Foreman's CI to stable branches on branching

### LINKED ISSUE #583: During branching, the version of katello used in tests should be pinned

### LINKED ISSUE #584: During branching, the packit config should be updated for Foreman
We did that for Katello in https://github.com/theforeman/theforeman-rel-eng/commit/7a5a851dc5f23cf05553566132b1b503ce18afb8
But not for Foreman, resulting in things like https://github.com/theforeman/foreman/pull/11003

ogajduse:
Pins Foreman's stable-branch CI to itself instead of develop/nightly:
- #583: pin the Katello version used in `foreman.yml`'s test job
- #584: update `.packit.yaml` on branching, like Katello already does

Builds on #447 (kept as a separate commit, credited to @ekohl). Since
Foreman and Katello now share the same X.Y version, this drops the
interactive Katello-version prompt in favor of `$VERSION`, and also
fixes the `.packit.yaml` `rpm/develop` wget URLs, which #447 missed
and which caused theforeman/foreman#11003.

Closes #583, closes #584.

### DIFF
--- a/branch_project
+++ b/branch_project
@@ -19,6 +19,20 @@ for repo in $(./ensure_clean_git_checkouts "$@") ; do
 			bundle exec librarian-puppet install
 			git add .gitignore Puppetfile*
 			git commit --message "Pin Puppet modules for ${VERSION}"
+		elif [[ $repo == foreman ]] ; then
+			# TODO: We don't know the Katello version at this point
+			read -p "Katello version: " -r KATELLO
+			if [[ -z $KATELLO ]] ; then
+				echo "No Katello version given" >&2
+				exit 1
+			fi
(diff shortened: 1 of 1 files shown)


zjhuntin (approved):
Ran locally and verified! LGTM.

ekohl:
This reminded me that you can also go in the other direction to simplify things, using the alignment: https://github.com/theforeman/theforeman-rel-eng/pull/608

